PR #18, opened 2022-01-02
Implemented Gate & City Chat

kbarkevich:
Gate & city chat implementation with option to change the colors of the text chat for each layer.

### DIFF
--- a/mh/pat.py
+++ b/mh/pat.py
@@ -1476,12 +1476,28 @@ class PatRequestHandler(SocketServer.StreamRequestHandler):
         The server sends a chat message.
         """
         data = struct.pack(">B", unk1)
-        info.text_color = pati.Long(0xbb3385ff)
-        info.sender_id = pati.String(b"C1I2D3")
-        info.sender_name = pati.String(b"Cid")
+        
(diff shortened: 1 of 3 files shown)


mh/pat.py
```diff
@@ -1476,12 +1476,28 @@ def sendNtcLayerChat(self, unk1, info, message, seq):
         The server sends a chat message.
         """
         data = struct.pack(">B", unk1)
-        info.text_color = pati.Long(0xbb3385ff)
-        info.sender_id = pati.String(b"C1I2D3")
-        info.sender_name = pati.String(b"Cid")
+
+        info.text_color = pati.Long(LAYER_CHAT_COLORS[self.session.layer])
+        info.sender_id = pati.String(self.session.capcom_id.encode("utf-8"))
```
InusualZ, line 1481: Actually, they use `ASCII` for the US/PAL and `SHIFT-JS` for the JAP game version. But since right know we don't even sanitize the input, it's fine. In any case the client would only read a certain amount of character (`capmcom_id` 0x08 and `hunter_name` 0x20) and doesn't complain about any non-supported character.

PR #19, opened 2022-01-02
Add a temporary database and some (un)packing utils

sepalani:
This PR adds (un)pacling utils and a temporary database.

Ready to be reviewed.


mh/database.py
```diff
@@ -0,0 +1,328 @@
...
+        for i in range(1, 3)
+    ])
+    servers.extend([
+        Server("Greed{}".format(i), ServerType.RECRUITING)
+        for i in range(1, 5)
+    ])
+    return servers
+
+
+class TempDatabase(object):
```
InusualZ, line 150: This class is suppose to be the interface of the persistent database that the server would have(?). I don't see much sense in saving the city or gates as part of this database since the players connected to the server, would need to be re-loaded/re-connected if a restart would come to happen.

mh/database.py
```diff
@@ -0,0 +1,328 @@
...
+        return len(self.players) + sum((
+            gate.get_population() for gate in self.gates
+        ))
+
+    def get_capacity(self):
+        return self.capacity
+
+
+def new_servers():
+    servers = []
```
InusualZ, line 130: Should we start loading and configuring the servers/gate/cities from a config file?

mh/database.py
```diff
@@ -0,0 +1,328 @@
...
+    FULL = 2
+
+
+class City(object):
+    def __init__(self, name, parent):
+        self.name = name
+        self.parent = parent
+        self.capacity = 4
+        self.state = LayerState.EMPTY
+        self.players = set()
```
InusualZ, line 57: What is your reasoning behind using `set` instead of `list`, to avoid duplicates more easily? I think the games have a few more packets with fields expecting a list index and with `set` I can't give them one because they are not reliable...

mh/database.py
```diff
@@ -0,0 +1,328 @@
...
+        session.capcom_id = capcom_id
+        session.hunter_name = name
+
+    def get_session(self, session):
+        """Returns existing PAT session or the current one."""
+        return self.sessions.get(session.pat_ticket, session)
+
+    def del_session(self, session):
+        """Delete the session from the database."""
+        # TODO: Find a good place to purge old tickets
```
InusualZ, line 224: About this comment, I think we are going to need some kind of scheduler or something a long those line to deal with this kind of task. Also, we are going to need it in order to simulate the day/night cycle with ticks

mh/session.py
```diff
@@ -0,0 +1,163 @@
...
+
+class Session(object):
+    """Server session class.
+
+    TODO:
+     - Finish the implementation
+    """
+    def __init__(self):
+        """Create a session object."""
+        self.local_info = {
```
InusualZ, line 38: Why create a dictionary for this variables? Instead of just plain vars?

mh/pat_item.py
```diff
@@ -79,6 +79,61 @@ def unpack_byte(data, offset=0):
     return value
 
 
+def pack_extra_info(info):
```
InusualZ, line 82: For this data struct name I was thinking something like `IndexSet` or `IndexedValues`, what do you thing?

mh/pat_item.py
```diff
@@ -325,6 +393,18 @@ def pack(self):
             for index, value in items
         )
 
+    def pack_fields(self, fields):
```
InusualZ, line 396: Such a elegant solution. Very Nice!

InusualZ (commented):
Mostly design comments

mh/database.py
sepalani: At some point, we'll need to synchronise the available servers if we use a decentralised model. The OPN, LMP and FMP servers are redirecting to the same address for all available servers which won't be the case when the FMP servers will be properly implemented. However, we'd need to keep track of the "registered" servers (for instance, multiple FMP servers).

I don't know if it's the best place to store that info but that's why for the time being I'm storing it in the TempDatabase.

mh/database.py
sepalani: Wouldn't be better to start this part when we have a proper db schema and the necessary API for inter-server communication? Currently, we share the state across server using the TempDatabase.

mh/database.py
sepalani: I was using a `set` to avoid duplicates but if it's causing issues I can switch it to a list.

mh/database.py
sepalani: I do agree, we'll have to tackle this issue, at some point.

mh/session.py
sepalani: I was too lazy to create a `LocalInfo` class that'd represent what the game calls "Local Info" in the online UI. It probably wouldn't be directly stored directly in the session but rather would be a link to another table when I was thinking in a SQL/database way.

mh/pat_item.py
sepalani: That's better than extra info, indeed. That gave me other naming idea, what about `ValueSet`, `FieldSet` or `optional_fields`? I'd tend to prefer `optional_fields` as the structure is similar to regular PAT's fields with an additional boolean flag.

mh/database.py
InusualZ: Ahhh, good point. I did not thought about it that way.

mh/database.py
InusualZ: Sound reasonable

mh/database.py
InusualZ: Yeah, I think it would be better to switch to `list` in any case. If we manage to introduce duplicated values on to the list, it would be a bug to fix

mh/session.py
InusualZ: That sound good

mh/pat_item.py
InusualZ: Indeed `optional_fields` is a good fit.

mh/database.py
sepalani: Alright, I'll amend this and update the PR, asap.

sepalani:
@InusualZ Done.

InusualZ:
LGTM

PR #24, opened 2022-01-11
PAT: Use layer's optional fields

sepalani:
This PR retains city's optional fields used to display the city properties at the gate.

I'll test it soon to make sure it doesn't break anything, meanwhile I'm setting it as a draft.


InusualZ (approved):
LGTM

PR #28, opened 2022-01-17
Added code for disconnecting from layers

kbarkevich:
… through means other than the city entrance

This does not include code for city owners to transfer ownership upon disconnection.

Edit: This now includes code for city owners to transfer ownership upon disconnection.

### DIFF
--- a/fmp_server.py
+++ b/fmp_server.py
@@ -300,17 +300,7 @@ class FmpRequestHandler(PatRequestHandler):
         Sent by the game when leaving the gate via the entrance:
          - Relocate > Select Server
         """
-        if self.session.layer == 2:
-            city = self.session.get_city()
-            if city.leader == self.session:
-                if len(city.players) > 1:
-                    # TODO: Transfer the leadership to another player
-                    pass
-                city.leader = None
-
-            ntc_data = pati.lp2_string(self.session.capcom_id)
-            self.server.layer_broadcast(self.session, PatID4.NtcLayerOut,
-                                        ntc_data, seq)
+        self.leave_layer()
         self.sendAnsLayerUp(data, seq)
 
     def recvReqUserSearchInfoMine(self, packet_id, data, seq):
--- a/fmp_server.py
+++ b/fmp_server.py
@@ -300,7 +300,7 @@ class FmpRequestHandler(PatRequestHandler):
         Sent by the game when leaving the gate via the entrance:
          - Relocate > Select Server
         """
-        self.leave_layer()
+        self.notify_layer_departure()
         self.sendAnsLayerUp(data, seq)
 
     def recvReqUserSearchInfoMine(self, packet_id, data, seq):
--- a/mh/constants.py
+++ b/mh/constants.py
@@ -236,6 +236,8 @@ def make_binary_trading_post():
 
 LAYER_CHAT_COLORS = (0xbb3385ff, 0xffffffff, 0xffffffff)
 
+LAYER_CHAT_COLORS = (0xbb3385ff, 0xffffffff, 0xffffffff)
+
 TERMS_VERSION = 1
 TERMS = {
     1: b"""MH3 Server Project - Terms.""",
--- a/mh/constants.py
+++ b/mh/constants.py
@@ -236,8 +236,6 @@ def make_binary_trading_post():
 
 LAYER_CHAT_COLORS = (0xbb3385ff, 0xffffffff, 0xffffffff)
 
-LAYER_CHAT_COLORS = (0xbb3385ff, 0xffffffff, 0xffffffff)
-
 TERMS_VERSION = 1
 TERMS = {
     1: b"""MH3 Server Project - Terms.""",
--- a/mh/pat.py
+++ b/mh/pat.py
@@ -246,7 +246,6 @@ class PatRequestHandler(SocketServer.StreamRequestHandler, object):
         The server sends a request to the game to establish a PAT connection.
         It also sends a parameter that seems unused on the western versions.
         """
-        self.session.graceful_shutdown = False
         data = struct.pack(">I", unused)
         self.send_packet(PatID4.ReqConnection, data, seq)
 
@@ -593,7 +592,7 @@ class PatRequestHandler(SocketServer.StreamRequestHandler, object):
         # 2: logout to different server(?)
         """
         shutdown_type, = struct.unpack(">B", data)
-        self.session.shutdown(shutdown_type)
+        self.session.attempt_leave_all_layers(shutdown_type)
         self.sendAnsShut(shutdown_type, seq)
 
     def sendAnsShut(self, shutdown_type, seq):
@@ -996,7 +995,7 @@ class PatRequestHandler(SocketServer.StreamRequestHandler, object):
         JP: レイヤ終了要求
         TR: Layer end request
         """
-        self.leave_layer()
+        self.notify_layer_departure()
         self.sendAnsLayerEnd(seq)
 
     def sendAnsLayerEnd(self, seq):
@@ -2375,7 +2374,7 @@ class PatRequestHandler(SocketServer.StreamRequestHandler, object):
         self.server.layer_broadcast(self.session, PatID4.NtcLayerHost, data,
                                     seq)
 
-    def leave_layer(self):
+    def notify_layer_departure(self):
         if self.session.layer == 2:
             new_host = self.session.try_transfer_city_leadership()
             if new_host:
@@ -2453,7 +2452,7 @@ class PatRequestHandler(SocketServer.StreamRequestHandler, object):
             self.session.disconnect()
             if not self.finished:
                 try:
-                    self.leave_layer()
+                    self.notify_layer_departure()
                 except Exception:
                     # We tried to comunicate the error to the client but it
                     # seems that the exception was caused by a connection
@@ -2461,7 +2460,7 @@ class PatRequestHandler(SocketServer.StreamRequestHandler, object):
                     # to communicate that error to the client
                     pass
 
-                self.session.shutdown(1)
+                self.session.attempt_leave_all_layers(1)
                 self.session.delete()
 
         self.server.del_from_debug(self)
--- a/mh/pat.py
+++ b/mh/pat.py
@@ -2439,18 +2439,20 @@ class PatRequestHandler(SocketServer.StreamRequestHandler, object):
         self.server.info("Handle client")
         self.server.add_to_debug(self)
         self.session = Session(self)
-
+        exception_raised = False
+        
         # Temporary workaround for Dolphin bug
         time.sleep(2)
         self.sendReqConnection()
         try:
             self.handle_client()
         except Exception as e:
+            exception_raised = True
             self.server.error(traceback.format_exc())
             self.send_error("{}: {}".format(type(e).__name__, str(e)))
         finally:
             self.session.disconnect()
-            if not self.finished:
+            if not self.finished or exception_raised:
                 try:
                     self.notify_layer_departure()
                 except Exception:
--- a/mh/pat.py
+++ b/mh/pat.py
@@ -246,6 +246,7 @@ class PatRequestHandler(SocketServer.StreamRequestHandler, object):
         The server sends a request to the game to establish a PAT connection.
         It also sends a parameter that seems unused on the western versions.
         """
+        self.session.graceful_shutdown = False
         data = struct.pack(">I", unused)
         self.send_packet(PatID4.ReqConnection, data, seq)
 
@@ -587,18 +588,22 @@ class PatRequestHandler(SocketServer.StreamRequestHandler, object):
         ID: 60100100
         JP: 切断要求
         TR: Disconnection request
-        """
-        login_type, = struct.unpack(">B", data)
-        self.sendAnsShut(login_type, seq)
 
-    def sendAnsShut(self, login_type, seq):
+        # 1: full logout(?)
+        # 2: logout to different server(?)
+        """
+        shutdown_type, = struct.unpack(">B", data)
+        self.session.shutdown(shutdown_type)
+        self.sendAnsShut(shutdown_type, seq)
+
+    def sendAnsShut(self, shutdown_type, seq):
         """AnsShut packet.
 
         ID: 60100200
         JP: 切断返答
         TR: Disconnection response
         """
-        data = struct.pack(">B", login_type)
+        data = struct.pack(">B", shutdown_type)
         self.send_packet(PatID4.AnsShut, data, seq)
         self.finish()
 
@@ -809,7 +814,7 @@ class PatRequestHandler(SocketServer.StreamRequestHandler, object):
         self.server.debug("UserObject: {}, {}, {!r}".format(
             is_slot_empty, slot_index, user_obj
         ))
-        hunter_name = ""
+        hunter_name = None
         if hasattr(user_obj, "hunter_name"):
             hunter_name = pati.unpack_string(user_obj.hunter_name)
         self.session.use_user(slot_index, hunter_name)
@@ -991,6 +996,7 @@ class PatRequestHandler(SocketServer.StreamRequestHandler, object):
         JP: レイヤ終了要求
         TR: Layer end request
         """
+        self.leave_layer()
         self.sendAnsLayerEnd(seq)
 
     def sendAnsLayerEnd(self, seq):
@@ -2357,6 +2363,28 @@ class PatRequestHandler(SocketServer.StreamRequestHandler, object):
         data = struct.pack(">B", success)
         self.send_packet(PatID4.AnsLayerMediationUnlock, data, seq)
 
+    def sendNtcLayerHost(self, new_leader, seq):
+        """NtcLayerHost packet.
+        ID: 64411000
+        JP: レイヤのホスト通知
+        TR: Layer host notification
+        """
+        data = struct.pack(">HII", 0x08, 0x00, 0x00)  # unk short array patitem
+        data += pati.lp2_string(new_leader.capcom_id)
+        data += pati.lp2_string(new_leader.hunter_name)
+        self.server.layer_broadcast(self.session, PatID4.NtcLayerHost, data,
+                                    seq)
+
+    def leave_layer(self):
+        if self.session.layer == 2:
+            new_host = self.session.try_transfer_city_leadership()
+            if new_host:
+                self.sendNtcLayerHost(new_host, 0)
+        if self.session.layer > 0:
+            ntc_data = pati.lp2_string(self.session.capcom_id)
+            self.server.layer_broadcast(self.session, PatID4.NtcLayerOut,
+                                        ntc_data, 0)
+
     def finish(self):
         self.finished = True
         try:
@@ -2418,11 +2446,23 @@ class PatRequestHandler(SocketServer.StreamRequestHandler, object):
         self.sendReqConnection()
         try:
             self.handle_client()
-            self.session.disconnect()
         except Exception as e:
             self.server.error(traceback.format_exc())
             self.send_error("{}: {}".format(type(e).__name__, str(e)))
-            self.session.delete()
+        finally:
+            self.session.disconnect()
+            if not self.finished:
+                try:
+                    self.leave_layer()
+                except Exception:
+                    # We tried to comunicate the error to the client but it
+                    # seems that the exception was caused by a connection
+                    # error and in that case there is nothing else we can do
+                    # to communicate that error to the client
+                    pass
+
+                self.session.shutdown(1)
+                self.session.delete()
 
         self.server.del_from_debug(self)
         self.server.info("Client finished!")
--- a/mh/session.py
+++ b/mh/session.py
@@ -231,9 +231,8 @@ class Session(object):
         return users[start:start+count]
 
     def leave_server(self):
-        old_server = self.local_info["server_id"]
-        if old_server is not None:
-            DB.leave_server(self, old_server)
+        if self.local_info["server_id"] is not None:
+            DB.leave_server(self, self.local_info["server_id"])
 
     def get_gates(self):
         return DB.get_gates(self.local_info["server_id"])
--- a/mh/session.py
+++ b/mh/session.py
@@ -233,7 +233,6 @@ class Session(object):
     def leave_server(self):
         old_server = self.local_info["server_id"]
         if old_server is not None:
-            self.local_info["server_id"] = None
             DB.leave_server(self, old_server)
 
     def get_gates(self):
--- a/mh/session.py
+++ b/mh/session.py
@@ -329,7 +329,7 @@ class Session(object):
         else:
             assert False, "Can't find layer"
 
-    def shutdown(self, shutdown_type):
+    def attempt_leave_all_layers(self, shutdown_type):
         assert 1 <= shutdown_type <= 2, "Invalid shutdown type"
         if shutdown_type != 1:
             return  # only shutdown_type 1 requires removal from layers
--- a/mh/session.py
+++ b/mh/session.py
@@ -150,6 +150,14 @@ class Session(object):
         return pati.getDummyLayerData()
 
     def layer_end(self):
+        if self.layer > 1:
+            # City path
+            self.leave_city()
+        if self.layer > 0:
+            # Gate path
+            self.leave_gate()
+            # Server path (executed at gate and higher)
+            self.leave_server()
         self.layer = 0
         self.state = SessionState.UNKNOWN
 
@@ -223,7 +231,10 @@ class Session(object):
         return users[start:start+count]
 
     def leave_server(self):
-        DB.leave_server(self)
+        old_server = self.local_info["server_id"]
+        if old_server is not None:
+            self.local_info["server_id"] = None
+            DB.leave_server(self, old_server)
 
     def get_gates(self):
         return DB.get_gates(self.local_info["server_id"])
@@ -256,6 +267,21 @@ class Session(object):
     def leave_city(self):
         DB.leave_city(self)
         self.state = SessionState.GATE
+
+    def try_transfer_city_leadership(self):
+        if self.local_info['city_id'] is None:
+            return None
+
+        city = self.get_city()
+        if city.leader != self:
+            return None
+
+        for _, player in city.players:
+            if player == self:
+                continue
+            city.leader = player
+            return player
+        return None
 
     def join_circle(self, circle_id):
         # TODO: Move this to the database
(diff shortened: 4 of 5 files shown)


mh/session.py
```diff
@@ -167,7 +167,9 @@ def get_layer_users(self, server_id, gate_id, city_id, first_index, count):
         return players[start:start+count]
 
     def leave_server(self):
-        DB.leave_server(self)
+        old_server = self.local_info["server_id"]
+        if old_server is not None:
```
InusualZ, line 171: I think that this is `if` is unnecessary because since you are using the `[]` operator, if the dictionary doesn't find the key it would raise a `KeyError` instead of returning `None`. I think that if you want to get a default value (`None`) you need to use the `get` function

fmp_server.py
```diff
@@ -305,18 +305,34 @@ def recvReqLayerUp(self, packet_id, data, seq):
...
+        TR: Layer end request
+        """
+        if self.session.layer > 1:
+            # City path
+            self.session.try_transfer_city_leadership()
+            ntc_data = pati.lp2_string(self.session.capcom_id)
+            self.server.layer_broadcast(self.session, PatID4.NtcLayerOut,
+                                        ntc_data, seq)
+            self.session.leave_city()
+        if self.session.layer > 0:
```
InusualZ, line 328: I think this should be made into a `elif`, so the layer is not compared twice in case of being `> 1`

InusualZ (changes requested):
Minor comments

mh/pat_item.py
```diff
@@ -131,6 +131,13 @@ def unpack_optional_fields(data, offset=0):
     return info
 
 
+def pack_unk_short_array_struct(capcom_id, name):
```
InusualZ, line 134: I don't think this should be made into a function, since this is only creating the payload for the `NtcLayerHost` packet.

It should be okay to create a function that assemble and send the packet. Something like 
```python
def sendNtcLayerHost(new_leader, seq):
    """NtcLayerHost packet.
    ID: 64411000
    JP: レイヤのホスト通知
    TR: Layer host notification
    """

    data = struct.pack(">HII", 0x08, 0x00, 0x00)  # unk short array pat item
    data += pati.lp2_string(new_leader.capcom_id)
    data += pati.lp2_string(new_leader.name)
    
    self.server.layer_broadcast(self.session, PatID4.NtcLayerHost, data, seq)

```
Also this function should be on `fmp_server.py`

fmp_server.py
InusualZ: This is actually the desired behavior (fallthrough)

mh/session.py
InusualZ: I'm wrong, when session is constructed a value (`None`) is assigned to that key

InusualZ (changes requested):

fmp_server.py
```diff
@@ -305,18 +305,51 @@ def recvReqLayerUp(self, packet_id, data, seq):
...
+            # Gate path
+            self.session.leave_gate()
+            # Server path (executed at gate and higher)
+            self.session.leave_server()
+
+        self.sendAnsLayerEnd(seq)
+    
+    def sendNtcLayerHost(self, new_leader, seq):
+        """NtcLayerHost packet.
+        ID: 64411000
```
sepalani, line 344: It's missing a newline here.

fmp_server.py
```diff
@@ -305,18 +305,51 @@ def recvReqLayerUp(self, packet_id, data, seq):
...
+            new_leader = self.session.try_transfer_city_leadership()
+            if new_leader is not None:
+                self.sendNtcLayerHost(new_leader, seq)
 
             ntc_data = pati.lp2_string(self.session.capcom_id)
             self.server.layer_broadcast(self.session, PatID4.NtcLayerOut,
                                         ntc_data, seq)
         self.sendAnsLayerUp(data, seq)
 
+    def recvReqLayerEnd(self, packet_id, data, seq):
```
sepalani, line 317: You're shadowing a method from the `pat` module, if you're sure that's only used by the FMP server and not the other ones, you should remove the implementation in the `pat` module. Otherwise, I'd suggest you move your code in the `pat` module instead.

mh/session.py
```diff
@@ -197,6 +199,16 @@ def join_city(self, city_id):
     def leave_city(self):
         DB.leave_city(self)
 
+    def try_transfer_city_leadership(self):
+        city = self.get_city()
```
sepalani, line 203: This method should perform the layer check and return `None` early if the `local_info[city_id]` is `None`.

fmp_server.py
```diff
@@ -305,18 +305,51 @@ def recvReqLayerUp(self, packet_id, data, seq):
...
+        """ReqLayerEnd packet.
+
+        ID: 64020100
+        JP: レイヤ終了要求
+        TR: Layer end request
+        """
+        if self.session.layer > 1:
```
sepalani, line 324: I'd like to avoid exposing the `layer` implementation logic to the server network code. A proper method should be created in `session` like `session.leave_layer()`.

Ideally, that'd reduce this method code to:
```python
new_host = self.session.try_transfer_city_leadership()
if new_host:
    self.sendNtcLayerHost(new_host, seq)
self.session.leave_layer()  # You might need to pass the server or self as parameter for the broadcast, I'll have to figure out a better design for this.
self.sendAnsLayerEnd(seq)
```
I'm also open to suggestions regarding other design possibilities.

fmp_server.py
```diff
@@ -305,18 +305,51 @@ def recvReqLayerUp(self, packet_id, data, seq):
          - Relocate > Select Server
         """
         if self.session.layer == 2:
-            city = self.session.get_city()
-            if city.leader == self.session:
-                if len(city.players) > 1:
-                    # TODO: Transfer the leadership to another player
-                    pass
-                city.leader = None
+            new_leader = self.session.try_transfer_city_leadership()
```
sepalani, line 308: I would have preferred a generic "layer" transfer of ownership something like `layer_migrate_host` but let's keep it like this for now.

mh/session.py
sepalani, line 171: When does an error occur? In which circumstance this check is needed?

sepalani (changes requested):

fmp_server.py
sepalani: You can also move the logic in the `session.layer_end` method if it's more fitting.

mh/session.py
kbarkevich: `session.get_city()` already does that check with an assertion. This function should only be able to be called from in the city, and when I programmed it I figured the `assert` within `get_city()` would be enough to cover that base. I should instead do an explicit `None` check within this function in addition to the check within `get_city()`?

fmp_server.py
kbarkevich: When I coded this I was valuing keeping the `layer_broadcast` in the server network code over keeping the `layer` implementation logic out of it. It came down to doing it one way or the other, since the logic of `layer_broadcast` (for both `NtcLayerHost` and `NtcLayerOut`) has to be done *before*  the server removes the player from the city, but *after* the server checks to see if the player is in a city.

I did pack a chunk of the logic into `session` already in the form of `try_transfer_city_ownership`, which notifies the `server` that packets need to be sent only if the layer changed hands to a new host.

So, I could move the logic into `session`, but I think it would require that I also move a lot of packet-sending logic in there as well, which there currently isn't any.

mh/session.py
sepalani: My point was to avoid the assert to be triggered because it will raise an exception that will close the connection. The advantage of this approach is that you don't need to check for the layer prior using this method. On top of that, it makes more sense for a function with `try` in the name not to fail.

fmp_server.py
sepalani: Feel free to disregard my point for now. I'll try to improve the current session abstraction. Basically, my initial thought was to separate the codebase like this:
 - session module: holds the current player network state.
 - database module: database abstraction interface.
 - server module: server module that will inherit protocol implementation or implement it.

I initially planned to implement the broadcast method into the session class but my implementation wasn't ready for that. So I used the server class instead.

mh/session.py
kbarkevich: Gotcha. This will be in the next commit.

mh/pat_item.py
kbarkevich: This has been fixed in a recent commit.

fmp_server.py
kbarkevich: Fixed in latest commit.

mh/session.py
kbarkevich: This will be necessary for the implementation of a graceful connection-closed handler. When a user terminates their connection (such as by continuing to press B all the way back to Mode Select) from within the server select screen, they might or might not already be connected to a server.

mh/session.py
```diff
@@ -200,13 +200,14 @@ def leave_city(self):
...
-        if city.leader == self:
-            if len(city.players) > 1:
-                for player in city.players:
-                    if player != self:
-                        city.leader = player
-                        return player
+        if self.local_info['city_id'] is not None:
+            city = self.get_city()
+            if city.leader == self:
+                if len(city.players) > 1:
```
sepalani, line 207: This size check seems unnecessary and is redundant with the `for` loop +`if player != self` check.

fmp_server.py
```diff
@@ -38,7 +38,22 @@ def recvAnsConnection(self, packet_id, data, seq):
         connection_data = pati.ConnectionData.unpack(data)
         self.server.debug("Connection: {!r}".format(connection_data))
         self.sendNtcLogin(3, connection_data, seq)
-
+    
+    def recvReqShut(self, packet_id, data, seq):
```
sepalani, line 42: This is shadowing its previous implementation and should be implemented in the `pat` module instead.

fmp_server.py
```diff
@@ -38,7 +38,22 @@ def recvAnsConnection(self, packet_id, data, seq):
...
+        """ReqShut packet.
+
+        ID: 60100100
+        JP: 切断要求
+        TR: Disconnection request
+        """
+        # 1: full logout(?)
+        # 2: logout to different server(?)
+        login_type, = struct.unpack(">B", data)
+        if login_type == 1 and self.session.layer == 0:
```
sepalani, line 52: This part shouldn't expose the `Session` class internal logic and should be implemented as a session method instead like `session.shutdown(login_type)`.

mh/pat.py
```diff
@@ -913,6 +913,33 @@ def recvReqLayerEnd(self, packet_id, data, seq):
...
+        if self.session.layer > 1:
+            # City path
+            new_leader = self.session.try_transfer_city_leadership()
+            if new_leader is not None:
+                self.sendNtcLayerHost(new_leader, seq)
+
+            ntc_data = pati.lp2_string(self.session.capcom_id)
+            self.server.layer_broadcast(self.session, PatID4.NtcLayerOut,
+                                        ntc_data, seq)
+            self.session.leave_city()
```
sepalani, line 925: IMHO, this whole `session.leave_` fallthrough should be done in the `session.layer_end` method.

mh/session.py
sepalani: I still don't think that's the right approach to add complexity to this function to address the `recvReqShut` case. `session.layer_end` is most-likely a better candidate as `session.leave_server` was intended to be only called when needed (which is why there are asserts in the functions it calls).

I'd be fine with something like this, if needed:
```python
DB.leave_server(self)
self.local_info["server_id"] = None
```

fmp_server.py
```diff
@@ -38,7 +38,22 @@ def recvAnsConnection(self, packet_id, data, seq):
...
-
+    
+    def recvReqShut(self, packet_id, data, seq):
+        """ReqShut packet.
+
+        ID: 60100100
+        JP: 切断要求
+        TR: Disconnection request
+        """
+        # 1: full logout(?)
```
sepalani, line 49: This can be added in the docstring directly.
```python
        """ReqShut packet.
        ID: 60100100
        JP: 切断要求
        TR: Disconnection request

        Login type:
         - 0x01: full logout(?)
         - 0x02: logout to different server(?)
        """
```

sepalani (changes requested):

fmp_server.py
kbarkevich: This is something that does need to be handled separately in `fmp_server`, since this is specifically an `fmp_server` `ReqShut` handler. `ReqShut` gets sent to each of the different servers at different times and for different reasons, so it still needs its own handler in the `pat` module, but this is (effectively) an `override` function for how we need to handle the `fmp_server` receiving it.

sepalani: I strongly believe that each shared packet has the same behaviour, if not there is a high chance that we're doing something wrong. That's not always true (like `recvAnsConnection`) but it often was.

What would not work if it's implemented in `pat` instead? What would be the missing data or the required one to make that works?

kbarkevich: I originally tried to implement it in `pat,` and realized that the other servers (the ones that are not `Fmp`) receive this exact same packet multiple times during the login process. While these do need to be responded to, thus the need for `pat` to have a simpler version of this implementation, the client sending this packet to `Fmp` specifically indicates a logging out of one of the game servers. This is the logout indicator that I was looking for that allowed me to remove the player from the game servers during a logout from the server select screen.

When I implemented it in `pat`, the result was the server experiencing an exception due to it removing players from the servers prematurely (during the login process.) One of the checks that I would have needed to do would have been to see that the server receiving it is `Fmp`, which is accomplished by using an override function in `fmp_server`.

mh/pat.py
kbarkevich: I'll implement it this way, but be aware that this will require some form of packet sending to be done in the `session` class. I can minimize this by passing them in as callback functions, but some form of sending logic will still need to be in `session` (such as the `seq` value now needing to be passed into `session.layer_end`)

mh/session.py
kbarkevich: Fixed in latest commit.

mh/pat.py
kbarkevich: I changed it in latest commit using callbacks, minimizing the amount of `session` logic in `fmp_server`. But let me know if you want me to switch it to what you've put here

fmp_server.py
kbarkevich: Fixed in latest commit.

mh/pat.py
sepalani: The callback idea is interesting but I think it shouldn't be used here. I'll keep that idea in mind as it will indeed be useful for implementing custom broadcast logic.

sepalani:
@kbarkevich @InusualZ  What do you think of this approach instead:

With this kind of logic it should work in the `pat` module directly.

```python
# In session.py

def layer_end(self):
    self.layer = 0

def shutdown(self, shutdown_type):
    assert 1 <= shutdown_type <= 2, "Invalid shutdown type"
    if shutdown_type != 1:
        return  # "soft" shutdown
    if self.local_info["city_id"]:
        self.leave_city()
    if self.local_info["gate_id"]:
        self.leave_gate()
    if self.local_info["server_id"]:
        self.leave_server()
    # I think there is a missing "self.local_info["server_id"] = None" in self.leave_server()
```

```python
# In pat.py
def recvReqShut(...):
    """(...)"""
    shutdown_type, = struct.unpack(">B", data)
    self.session.shutdown(shutdown_type)
    self.sendAnsShut(shutdown_type, seq)

def recvReqLayerEnd(...):
    """(...)"""
    new_host = self.session.try_transfer_city_leadership()
    if new_host:
        self.sendNtcLayerHost(new_host, seq)
    
    ntc_data = pati.lp2_string(self.session.capcom_id)
    self.server.layer_broadcast(self.session, PatID4.NtcLayerOut,
                                ntc_data, seq)
    self.sendAnsLayerEnd(seq)
```

InusualZ:
I haven't being following the discussion very closely, but from the changes that I can see and some of the comments. I would agree with @sepalani . Also, I don't like that callback approach tbh.

I like your logic @sepalani, but were would `layer_end()` would be used? On `recvReqLayerEnd`?

sepalani:
> I haven't being following the discussion very closely, but from the changes that I can see and some of the comments. I would agree with @sepalani . Also, I don't like that callback approach tbh.
> 
> I like your logic @sepalani, but were would `layer_end()` would be used? On `recvReqLayerEnd`?

IIRC, it's already used in `sendAnsLayerEnd` and implemented.

kbarkevich:
> ```python
> def recvReqLayerEnd(...):
>     """(...)"""
>     new_host = self.session.try_transfer_city_leadership()
>     if new_host:
>         self.sendNtcLayerHost(new_host, seq)
>     
>     ntc_data = pati.lp2_string(self.session.capcom_id)
>     self.server.layer_broadcast(self.session, PatID4.NtcLayerOut,
>                                 ntc_data, seq)
>     self.sendAnsLayerEnd(seq)
> ```

I believe this would work, but I would need to put some session logic (checking to see if the player is in the city layer) to this method before checking for a new host and then broadcasting the `NtcLayerOut` packet. Then I can try implementing it this way.

kbarkevich:
The latest commit now has these changes. I tested it and it still seems to work in all cases.

mh/pat.py
kbarkevich: Callback was removed in recent commit and changed to recommended logic.

InusualZ:
Haven't tested it, but code LGTM

mh/session.py
```diff
@@ -54,6 +54,7 @@ def __init__(self):
         self.binary_setting = b""
         self.search_payload = None
         self.hunter_info = pati.HunterSettings()
+        self.graceful_shutdown = False
```
sepalani, line 57: This can be avoided since it only happens when an exception occurs. You just need to clean it in the `except` block.

Example:
```python
    def handle(self):
        """Default PAT handler."""
        (...)
        self.sendReqConnection()
        try:
            self.handle_client()
        except Exception as e:
            traceback.print_exc()
            # Insert the cleaning logic here
            # self.leave_layer() ???
```

mh/session.py
sepalani: This part still seems superfluous to me:
```python
old_server = self.local_info["server_id"]
if old_server is not None:
```

fmp_server.py
```diff
@@ -898,6 +906,19 @@ def sendAnsCircleMatchEnd(self, seq):
...
+            if self.session.layer == 2:
+                new_host = self.session.try_transfer_city_leadership()
+                if new_host:
+                    self.sendNtcLayerHost(new_host, 0)
+            if self.session.layer > 0:
+                ntc_data = pati.lp2_string(self.session.capcom_id)
+                self.server.layer_broadcast(self.session, PatID4.NtcLayerOut,
+                                            ntc_data, 0)
+            self.session.shutdown(1)
+        self.session.graceful_shutdown = False
```
sepalani, line 921: I don't think it's useful as it will be set to `False` again on the next connection.

mh/session.py
```diff
@@ -215,3 +237,15 @@ def get_layer_players(self):
             return city.players
         else:
             assert False, "Can't find layer"
+            
+    def shutdown(self, shutdown_type):
+        assert 1 <= shutdown_type <= 2, "Invalid shutdown type"
+        self.graceful_shutdown = True
+        if shutdown_type != 1:
+            return  # only shutdown_type 1 requires removal from layers
+        if self.local_info["city_id"]:
```
sepalani, line 246: This logic is shared with `session.layer_end`, it's probably better to move it to its own function like `clear_local_info`.

sepalani (changes requested):
Some code logic and deduplication nitpicks.

fmp_server.py
kbarkevich: I thought so too, and it took me a little bit to figure this out, but it breaks if this line is not there. I can only assume that the "master" PatHandler class that this project inherits from has some underlying logic that inherits the old `Session`.

sepalani: Ah, that's right. Rather than implementing it to the `ReqConnection` packet, you should instead implement it in the `session.get` method instead. Though, as I've mentioned, I'm pretty sure it can be dealt with by cleaning it up in the `except` block of the `handle` method.

sepalani: In my local wip branch I changed this part: https://github.com/sepalani/MH3SP/blob/703cc1c086a6e181ed474776fb00c6430eaa4cc0/mh/session.py#L69

I added this guard:
```python
session = DB.get_session(self)
if session != self:
    assert session.connection is None, "Session is already in use"
    session.connection = self.connection
    self.connection = None
return session
```

fmp_server.py
kbarkevich: Only *just* realized what was missing the newline here. It will be fixed in the next commit.

mh/pat.py
```diff
@@ -246,6 +246,7 @@ def sendReqConnection(self, unused=0, seq=0):
         The server sends a request to the game to establish a PAT connection.
         It also sends a parameter that seems unused on the western versions.
         """
```
InusualZ, line 248: Variable no longer needed.

mh/pat.py
```diff
@@ -2357,6 +2363,28 @@ def sendAnsLayerMediationUnlock(self, index, mediation_data, seq):
...
+        JP: レイヤのホスト通知
+        TR: Layer host notification
+        """
+        data = struct.pack(">HII", 0x08, 0x00, 0x00)  # unk short array patitem
+        data += pati.lp2_string(new_leader.capcom_id)
+        data += pati.lp2_string(new_leader.hunter_name)
+        self.server.layer_broadcast(self.session, PatID4.NtcLayerHost, data,
+                                    seq)
+
+    def leave_layer(self):
```
InusualZ, line 2378: ‘leave_layer’ is not a proper name for what this function actually does, maybe a better name could be ‘notify_layer_departure’?

mh/pat.py
```diff
@@ -2413,14 +2441,27 @@ def handle(self):
         self.server.add_to_debug(self)
         self.session = Session(self)
 
+        time.sleep(2)
```
InusualZ, line 2444: I know it’s not your fault but could you add back the comment saying that it’s a workaround for a dolphin bug

mh/session.py
```diff
@@ -303,6 +329,17 @@ def get_layer_players(self):
         else:
             assert False, "Can't find layer"
 
+    def shutdown(self, shutdown_type):
```
InusualZ, line 332: This should be named something more appropriate to what the function is doing, maybe ‘full_layers_end’?

mh/constants.py
```diff
@@ -234,6 +234,8 @@ def slot(item, qty):
     return data
 
 
+LAYER_CHAT_COLORS = (0xbb3385ff, 0xffffffff, 0xffffffff)
```
InusualZ, line 237: Duplicated definition of variable

InusualZ (changes requested):
Overall looks good. Some cosmetics changes and we should be good to go.

sepalani (changes requested):
The logic sounds sane to me. I haven't tested it yet.

mh/pat.py
```diff
@@ -2412,15 +2439,31 @@ def handle(self):
         self.server.info("Handle client")
         self.server.add_to_debug(self)
         self.session = Session(self)
-
+        exception_raised = False
+        
+        # Temporary workaround for Dolphin bug
+        time.sleep(2)
```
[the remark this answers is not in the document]
InusualZ: Yes

mh/pat.py
```diff
@@ -2412,15 +2439,31 @@ def handle(self):
...
+            self.session.disconnect()
+            if not self.finished or exception_raised:
+                try:
+                    self.notify_layer_departure()
+                except Exception:
+                    # We tried to comunicate the error to the client but it
+                    # seems that the exception was caused by a connection
+                    # error and in that case there is nothing else we can do
+                    # to communicate that error to the client
+                    pass
```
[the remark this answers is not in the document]
InusualZ: We need the `exception_raise` var, because if a non socket related exception is catch, `self.finished` would be `False` but when the exception message is sent to the client through the `sendNtcShut` packet (which is called through `self.send_error`) the `self.finish` method would also be called which would set `self.finished = True` which would make this:
```py
if not self.finished:
(...)
```

not execute because `self.finished == False`, but it's a valid path because we need to correctly remove the player from the layers. 

I do agree about not needing the `finally` block and about calling `session.layer_end` inside the `session.delete` method

mh/pat.py
```diff
@@ -587,18 +587,22 @@ def recvReqShut(self, packet_id, data, seq):
...
         JP: 切断要求
         TR: Disconnection request
+
+        # 1: full logout(?)
+        # 2: logout to different server(?)
         """
-        login_type, = struct.unpack(">B", data)
-        self.sendAnsShut(login_type, seq)
+        shutdown_type, = struct.unpack(">B", data)
+        self.session.attempt_leave_all_layers(shutdown_type)
```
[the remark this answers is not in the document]
kbarkevich: self.session.layer_end() only removes you from the server (both session-wise and in the DB) if you are in layer 1 or above. attempt_leave_all_layers removes you from all layers regardless of where you are.

mh/session.py
```diff
@@ -223,7 +231,10 @@ def find_users(self, capcom_id, hunter_name, first_index, count):
         return users[start:start+count]
 
     def leave_server(self):
-        DB.leave_server(self)
+        old_server = self.local_info["server_id"]
+        if old_server is not None:
+            self.local_info["server_id"] = None
```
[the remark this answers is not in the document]
kbarkevich: Fixed

mh/pat.py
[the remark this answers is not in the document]
InusualZ: @kbarkevich You missed this comment

InusualZ (changes requested):

sepalani (changes requested):
You should also squash your commits into a single one.

`git rebase -i` and `git commit --amend`, might help you to keep the commit history cleaner.

mh/pat.py
```diff
@@ -2412,15 +2439,25 @@ def handle(self):
...
         except Exception as e:
+            exception_raised = True
             self.server.error(traceback.format_exc())
             self.send_error("{}: {}".format(type(e).__name__, str(e)))
-            self.session.delete()
+
+        self.session.disconnect()  # Always disassociate capcom_id with current session
+        if not self.finished or exception_raised:  # Abnormal disconnection, clean up
+            try:
+                self.notify_layer_departure()
```
sepalani, line 2457: It's missing an `except Exception as e:` block. Without it, the rest of the code after the `finally` block won't be executed.

sepalani (changes requested):

PR #29, opened 2022-01-25
Fix CapcomID persistence across session

sepalani:
This PR prevents multiple users to use simultaneously the same Capcom ID. It also provides a wrapper that can forward tiny error messages to the game.

### DIFF
--- a/mh/session.py
+++ b/mh/session.py
@@ -65,7 +65,7 @@ class Session(object):
             self.online_support_code = to_str(
                 pati.unpack_string(connection_data.online_support_code)
             )
-        session = DB.get_session(self)
+        session = DB.get_session(self.pat_ticket) or self
         if session != self:
             assert session.connection is None, "Session is already in use"
             session.connection = self.connection
--- a/mh/session.py
+++ b/mh/session.py
@@ -33,7 +33,7 @@ class Session(object):
     TODO:
      - Finish the implementation
     """
-    def __init__(self):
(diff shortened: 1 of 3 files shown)


mh/session.py
```diff
@@ -65,15 +65,33 @@ def get(self, connection_data):
             self.online_support_code = to_str(
                 pati.unpack_string(connection_data.online_support_code)
             )
-        return DB.get_session(self)
+        session = DB.get_session(self)
+        if session != self:
```
InusualZ, line 69: This code is a little bit confusing, you are inside a class named `Session` and asking a database about getting session... Once I looked a bit thorough, I understand what was happening. Maybe giving a new name to `DB.get_sesssion()`?

InusualZ (approved):
Just one comment, other than that code LGTM

mh/session.py
sepalani: I think finding a new name will be hard for that case _(sessions which aren't connected yet)_. What about changing its parameter instead?

Would something like this sounds fine to you:
```
session = DB.get_session(pat_ticket) or self
```
Or alternatively:
```
session = DB.get_session(pat_ticket, default=self)
```

InusualZ: I think the first option give a little more about the intent of the function.

sepalani: Done in the last commit.

InusualZ:
LGTM

PR #30, opened 2022-01-26
PAT: Check that the handler (still) exists before a broadcast

sepalani:
This PR fixes a bug when, in some circumstances like disconnection or race condition, the `pat_handler` is `None` and try to call the `send_packet` method.

### DIFF
--- a/mh/pat.py
+++ b/mh/pat.py
@@ -79,8 +79,9 @@ class PatServer(SocketServer.ThreadingTCPServer, Logger):
             if exclude_self and player == session:
                 continue
 
-            pat_handler = self.get_pat_handler(player)
-            pat_handler.send_packet(packet_id, data, seq)
+            handler = self.get_pat_handler(player)
+            if handler:
+                handler.try_send_packet(packet_id, data, seq)
 
     def circle_broadcast(self, circle, packet_id, data, seq,
                          session=None):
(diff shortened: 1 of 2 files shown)


InusualZ:
I thought about doing this, but I thought that in the cases that we weren't receiving a proper `pat_handler` was because we weren't clearing the session from the layers properly. So, I thought that #28 would fix the issue

sepalani:
That's one of the case but I'm pretty sure it can happen with a race condition too. Since there are 2 threads running in parallel, the one from the session which is going to be deleted and the one who wants to perform a broadcast.

However, now that I think about it, this simple approach won't work because network errors can occur too and we don't want a partial broadcast failure to kick the player. I'll give it some thought. I have different ideas in mind like:
 * Wrapping it in a `try..except` block _(quick and dirty approach, IMO)_
 * Adding a queue system somewhere in the server thread (not the player one)
   - Maybe during the `select` loop.
   - Need more thought and might be slower, too

I'll turn it into a draft for now.

sepalani:
Rather than creating a wrapper class, I created a proper method that handles exception. This should avoid to use wrapper or any magic tricks that can make the code harder to follow.

mh/pat.py
```diff
@@ -134,6 +136,17 @@ def send_packet(self, packet_id=0, data=b'', seq=0):
             packet_id, seq, hexdump(data)
         )
 
+    def try_send_packet(self, packet_id=0, data=b'', seq=0):
+        """Send PAT packet and catch exceptions."""
+        try:
+            self.send_packet(packet_id, data, seq)
+        except Exception as e:
+            self.server.warning(
```
InusualZ, line 144: I think a more useful information would be the exception message. Having the packet info doesn’t tell me anything what really happened.

InusualZ (changes requested):
Logic look sane, 1 comment though.

sepalani:
@InusualZ 
Done. I added the traceback in the log message.

PR #32, opened 2022-02-06
Database: Fix Capcom ID creation

sepalani:
This PR fixes an assertion error which spotted an issue in the Capcom ID creation. This assertion happens when you select an empty user slot.

Ready to be reviewed & merged.


InusualZ:
LGTM

PR #36, opened 2022-03-02
PAT: Implement LayerDetailSearch packets

sepalani:
This PR implements the city search feature that can be used at the gate next to the arena search lady.

### DIFF
--- a/mh/database.py
+++ b/mh/database.py
@@ -105,6 +105,8 @@ class Circle(object):
 
 
 class City(object):
+    LAYER_DEPTH = 3
+
     def __init__(self, name, parent):
         self.name = name
         self.parent = parent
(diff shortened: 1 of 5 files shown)


InusualZ (commented):
Implementation is fairly similar, I like it :)

I tested the change but for some reason I wasn't able to join the city from the city search dialog
image
The client send 2 `NtcCollectionLog` packet
image

It's up to you if you want to fix that in this PR. In either case LGTM

sepalani:
Joining a city from any kind of warp (city search, gate change, friend list, etc.) doesn't work and isn't implemented by this PR. BTW, I haven't mentioned it but the arena search lady is also working since it's using the same packet.

IIRC, the issue with NtcCollectionLog is tied to the NetworkUniqId that doesn't seem to be valid. Clearly something that we'll have to investigate at some point.

PR #38, opened 2022-04-01
PAT: Add PatData support to the Unpacker class

sepalani:
This PR adds PatData support to the Unpacker class. That way, the offset computations are done automatically by the unpacker on top of checking the packing/unpacking consistency.

### DIFF
--- a/mh/pat_item.py
+++ b/mh/pat_item.py
@@ -773,12 +773,20 @@ class LayerBinaryInfo(PatData):
     )
 
 
-class Unpacker(GenericUnpacker):
-    """Simple PAT item unpacker.
+def patdata_extender(unpacker):
+    """PatData classes must be defined above this function."""
+    for name, value in globals().items():
+        if not isinstance(value, type):
+            continue  # Not a valid class
+        if not issubclass(value, PatData):
+            continue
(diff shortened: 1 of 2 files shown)


mh/pat_item.py
```diff
@@ -773,12 +773,20 @@ class LayerBinaryInfo(PatData):
     )
 
 
+def patdata_extender(unpacker):
+    """PatData classes must be defined above this function."""
+    for name, value in globals().items():
+        if not isinstance(value, type):
```
InusualZ, line 779: I'm sorry if it's obvious but from where `type` come from?

sepalani: It's a kinda obscure concept in python. `type` is a class. This class is used for debugging and meta-programming. `type` returns the type of an object and can also be used to create a class on-the-fly programmatically.

`type` used to check an object's type:
image


Here I need to check that the value is a valid python class (otherwise `issubclass` will throw an exception):
 - https://stackoverflow.com/a/395741

InusualZ: Ahhhhh, I see

InusualZ (approved):
LGTM

PR #39, opened 2022-04-10
Replace `Players` collection

InusualZ:
The game uses a collection of slot to store the players of the current layer. In some packets the server must send a valid player index and this index must reflect the position in the slot collection across all players in the same layer. With our previous collection (plain `list`) this wasn't possible because if at some point someone was removed from the collection, all other player index position would become invalid.

Now we implement that sort of collection with a list of slot (default `None`). When someone is added we give it the first empty (`None`) slot and when someone is removed we empty-out (set to `None`) the slot. With this any other player index position stay the same throughout their connection in the layer.

### DIFF
--- a/fmp_server.py
+++ b/fmp_server.py
@@ -19,6 +19,7 @@
     along with this program.  If not, see <http://www.gnu.org/licenses/>.
 """
 
+from mh.database import Players
 import mh.pat_item as pati
 from mh.constants import *
 from mh.pat import PatRequestHandler, PatServer
@@ -141,7 +142,7 @@ class FmpRequestHandler(PatRequestHandler):
         count = len(players)
         data = struct.pack(">I", count)
 
-        for player in players:
+        for _, player in players:
             user = pati.LayerUserInfo()
             user.capcom_id = pati.String(player.capcom_id)
             user.hunter_name = pati.String(player.hunter_name)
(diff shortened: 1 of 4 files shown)


fmp_server.py
```diff
@@ -530,13 +530,11 @@ def recvReqCircleJoin(self, packet_id, data, seq):
             self.sendAnsCircleJoin(0, 0, seq)
             return
 
-        # TODO: When appending get the index of the slot
-        circle.players.append(self.session)
+        player_index = circle.players.add(self.session)
+        assert player_index != -1, "City is full"
```
sepalani, line 534: Shouldn't it be `Circle is full` rather than `City`?

sepalani (commented):
LGTM code-wise. I'm going to test it.

sepalani:
After some testing, some exceptions are raised and need to be addressed for some packets:
 - [AnsLayerDetailSearchData](https://github.com/sepalani/MH3SP/blob/f1eb7ee1c962cd02eed43cead9ef8bfe1b2214de/mh/pat.py#L2117), `Players object doesn't support indexing`
 - [AnsUserSearchData](https://github.com/sepalani/MH3SP/blob/f1eb7ee1c962cd02eed43cead9ef8bfe1b2214de/mh/pat.py#L1166), `Session object is not iterable`

Do the packets related to the search features need this slot index? If not, leaving the search result as a list of sessions should be fine if there are no overhead.

InusualZ:
Ahh I forgot to test after rebasing onto `dev`. Let me take a look.

InusualZ:
> Do the packets related to the search features need this slot index? If not, leaving the search result as a list of sessions should be fine if there are no overhead.

To be honest, I haven't look into that, that much but I don't think the use this slot index. Since that slot index is mostly used when you are inside the city/circle and since the searching functionality don't actually interact with those layer I don't think they would ever use them.

InusualZ:
Oops forgot to attach screenshot taken from the tests

*User Search*
image

*City Search*
image

sepalani (approved):
LGTM

PR #43, opened 2022-04-21
Track the session state

### LINKED ISSUE #41: [dev] The quest's host can force party member to depart for a quest.
Party member don't need to be ready before departure.

InusualZ:
We needed to track the session (player) a little bit with more detail, so I came up with this idea. But to be honest, after implementing it, I'm not too exited about it. Maybe there is better/simpler way?

Anyway, this allow me to fix two things:
- Previously, if someone created a circle and someone joined the circle and set himself as ready (standby), but another player joined after him, that other player wouldn't be able to see him (when in-quest) because the second player was ready (standby) before the third joined the circle. In other words, we weren't communicating the players that were ready, before you joined.
- Previously, we were sending to the circle's quest, everyone who enlisted in the circle ignoring if they were ready (standby) or not.

fix #41

### DIFF
--- a/fmp_server.py
+++ b/fmp_server.py
@@ -851,13 +851,24 @@ class FmpRequestHandler(PatRequestHandler):
         circle = self.session.get_circle()
         circle.departed = True
 
-        count = circle.get_population()
-        data = struct.pack(">I", count)
+        count = 0
+        data = b''
         for i, player in circle.players:
-            data += struct.pack(">B", i+1)
-            data += pati.lp2_string(player.capcom_id)
-            data += pati.lp2_string(b"\1")
-            data += struct.pack(">H", 21)  # TODO: Field??
+            if player.is_circle_standby():
+                data += struct.pack(">B", i+1)
+                data += pati.lp2_string(player.capcom_id)
+                data += pati.lp2_string(b"\1")
+                data += struct.pack(">H", 21)  # TODO: Field??
+                count += 1
+            else:
+                # Client ignore field
+                ntc_circle_kick = struct.pack(">B", 0) + pati.lp2_string('')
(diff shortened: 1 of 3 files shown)


fmp_server.py
```diff
@@ -846,13 +851,23 @@ def sendNtcCircleMatchStart(self, seq):
...
+                data += struct.pack(">B", i+1)
+                data += pati.lp2_string(player.capcom_id)
+                data += pati.lp2_string(b"\1")
+                data += struct.pack(">H", 21)  # TODO: Field??
+                count += 1
+            else:
+                # Client ignore field
+                ntc_circle_kick = struct.pack(">B", 0) + pati.lp2_string('')
+                pat_handler = self.server.get_pat_handler(player)
+                pat_handler.send_packet(PatID4.NtcCircleKick, ntc_circle_kick,
```
sepalani, line 867: I think the bug I mentioned resides here. We should probably remove the kicked player from the circle.

Maybe doing something like this would solve the issue:
```python
kick_list = []
for i, player in circle.players:
    if not player.is_circle_standby():
        kick_list.append(player)
        continue
    data +=  # Rest of the code
    # (...)

ntc_circle_kick = struct.pack(">B", 0) + pati.lp2_string('')
for player in kick_list:
    pat_handler = self.server.get_pat_handler(player)
    pat_handler.send_packet(PatID4.NtcCircleKick, ntc_circle_kick, seq)
    circle.players.remove(player)
```

InusualZ: Yep, you are right. I forgot to remove the player from the circle's player collection.

So I modified the `Players::remove` method so it can accept the index, for a more direct removal of the player.

sepalani:
LGTM

PR #45, opened 2022-04-30
Circle password

InusualZ:
Previously, when leader of a circle (quest) set a password, we weren't verifying it when another player joined the circle.

image

Also, we now have a proper method of communicating with the client a non-fatal error message


sepalani:
Tested and LGTM.

PR #46, opened 2022-04-30
Fix player location in city's player list

InusualZ:
Fix player location in city's player list

In City|Quest
-|-
image|image

Fix player location when searching city

In City|Quest
-|-
image|image


sepalani:
Tested and LGTM.

PR #47, opened 2022-05-03
Properly propagate changes in the circle's member collection

InusualZ:
We now have the capability to properly serialize our self the member of a circle. That way when we made changes to the collection of participant of a circle, we can send the changes without waiting for the circle's leader to acknowledge the changes himself.

For now, we only fix the case were if somebody is left behind (because he/she wasn't ready) when departing for a quest, we propagate the change correctly.

Other fixed would come after we properly detect and cleanup when somebody disconnect


sepalani:
Tested & LGTM.

PR #50, opened 2022-08-16
Properly shutdown when requested

InusualZ:
Previously weren't shutting down the `StreamRequestHandler` when sending `AnsShut`/`NtcShut`. This caused problems because when the handler was going to loop back and `::select` the socket, it was closed by the client (requested by us) causing the system to throw an exception that wouldn't be handled, This in-turn caused that the `session`'s resources weren't being saved/free properly.


sepalani (approved):
LGTM.

PR #51, opened 2022-08-19
Load balancing packet handling

InusualZ:
The idea behind load balancing the handling of packets is the following:
- The server and his clients would be serviced in the same thread.
  - By serviced I mean that we would wait for an event (*mainly read*) to happen for those sockets in the same thread. This because the game have an inherent 640 players cap (US/PAL) so at most the server thread would be managing 641 sockets (640 clients and the server) which I think most server hardware or modern PC should be able to handle. Even more so since that thread would only be responsible for copying the bytes from the socket internal buffer to the software buffer.
- After the packet have been received, they would be distributed to worker threads for actual handling of the payload.
  - Since this expected to be the most expensive part of this operation, I decided to distribute the load across multiple threads. 
  
### Things left for future PR
With this approach we need to think about data race and synchronizations, since data may be accessed from multiple thread at the same time. We haven't being experiencing this issues because in the environment were this changes have been tested (Closed Beta) the hardware has only one 1 CPU Core and the amount of worker threads is based on the amount of CPU Core.


mh/pat.py
```diff
@@ -2398,26 +2376,82 @@ def sendNtcLayerHost(self, new_leader, seq):
...
 
+        self.session.layer_end()
+        self.session.disconnect()
+        self.session.delete()
+
+        self.server.del_from_debug(self)
+        self.server.info("Client finished!")
+
+    def on_packet(self, data):
+        if data is None:
```
sepalani, line 2435:
```python
if not data:
```

mh/pat.py
```diff
@@ -2398,26 +2376,82 @@ def sendNtcLayerHost(self, new_leader, seq):
...
+        )
+        self.dispatch(packet_id, data, seq)
+
+    def send_packet(self, packet_id=0, data=b'', seq=0):
+        super(PatRequestHandler, self).send_packet(packet_id, data, seq)
+        self.server.debug(
+            "SEND %s[ID=%08x; Seq=%04x]",
+            PAT_NAMES.get(packet_id, "Packet"),
+            packet_id, seq
+        )
```
sepalani, line 2453: You aren't logging packet data anymore which could be crucial in some debugging/reverse engineering case. What about:
 - switching the current default log level to INFO
 - use `self.server.info` instead of `self.server.debug`
 - add the `hexdump(data)` of the packet in a `self.server.debug` call?

mh/basic.py
```diff
@@ -0,0 +1,374 @@
...
+    def on_tick(self):
+        """Called every time the server tick
+
+        ** Currently executed from the server thread
+
+        """
+        pass
+
+    def on_finish(self):
+        """Called beforte finish"""
```
sepalani, line 123: `before`

mh/basic.py
```diff
@@ -0,0 +1,374 @@
...
+from time_utils import Timer
+
+try:
+    # Python 3
+    import selectors
+    import queue
+except ImportError:
+    # Python 2
+    import selectors2 as selectors
+    import Queue as queue
```
sepalani, line 38: Do you mind refactoring the Server class so that it has a "dumb" implementation using `select` and a `SelectorMixIn` _(a bit like the `ForkingMixIn` and `ThreadingMixIn` class)_? That way the code architecture doesn't depend on this dependency and it doesn't force regular _(python2)_ users to download an external dependency. You could then create the `SelectorMixIn` class only if this dependency is available.

mh/basic.py
```diff
@@ -0,0 +1,374 @@
...
+
+        """
+        self.__shutdown_request = True
+
+        if sys.platform != "win32":
+            # Non Windows selector don't support waking up
+            # the select when a socket is closed, to bypass
+            # that we can just create a new socket that connect
+            # to the server, that way it would wake up the select
+            # and then see that it's closed and be done with it
```
sepalani, line 347: Do you have any reference/link to this? That sounds like a really strange bug.

mh/basic.py
```diff
@@ -0,0 +1,374 @@
...
+            try:
+                tmp = socket.socket(self.address_family, self.socket_type)
+                tmp.connect(self.server_address)
+                tmp.close()
+            except Exception as e:
+                print('Unable to close server. {}'.format(e))
+                pass
```
sepalani, line 354: `pass` isn't needed here.

mh/pat.py
InusualZ: Great idea

mh/basic.py
InusualZ: This is actually not needed anymore, since we are waking up at least once every time the server thread's `select` timeout (1 second). So, at worst we would be waiting here 1 second before the server thread wake up and set `self.__is_shut_down`

mh/basic.py
InusualZ: I don't know if you still want this in? I made the change that we agreed upon on discord, which was to include the `selectors2` dependency in the project since it's a single file with all the implementation (`externals/selectors2.py`).

sepalani: Could you please add another directory so that the path is `externals/selectors2/selectors2.py` so we can add the LICENSE file in that folder too. This structure will also allow us to use more complex dependencies using submodules if that's really needed.

Otherwise, I'm still OK for it to be included, as you said the MIT licence doesn't conflict with the project's AGPL licence.

InusualZ: Done

other/utils.py
```diff
@@ -392,7 +395,6 @@ def server_main(name, server_class, server_handler):
             thread.join(0.1)  # Timeout allows main thread to handle signals
             if args.log_to_window:
                 ui_update()
-    except KeyboardInterrupt:
+    except (KeyboardInterrupt, Exception):
         server.info("Interrupt key was pressed, closing server...")
-        server.shutdown()
-        server.server_close()
+        server.close()
```
sepalani, line 400: What about adding another `except Exception` block and logging the exception via `server.error` then with a `finally` statement call `server.close`?
```python
try:
    # (...)
except KeyboardInterrupt:
    server.info("Interrupt key was pressed, closing server...")
except Exception:
    server.error(...)
finally:
    server.close()
```

master_server.py
```diff
@@ -78,11 +78,10 @@ def interactive_mode(local=locals()):
 
         if args.interactive:
             t.join()
-    except KeyboardInterrupt:
+    except (KeyboardInterrupt, Exception):
         print("Interrupt key was pressed, closing server...")
         for server in servers:
-            server.shutdown()
-            server.server_close()
+            server.close()
```
sepalani, line 84: Ditto for the `except Exception`+`finally`.

sepalani (changes requested):
I'm going to test it thoroughly next Wednesday. If I don't encounter issues I'll be merging it.

sepalani:
This PR is working as expected beside one thing. If you run twice a server, the second one won't error nor print an exception which is unusual.

@InusualZ 
I'm fine with the current state of this PR. So feel free to merge it as is.

sepalani:
PR rebased to allow testing with python3. The following error was raised with python3 during my testing:
```
[2022-10-28 14:23:43 | OpnServer] Traceback (most recent call last):
  File "C:\Users\sepal\Projects\MHTri Project\MH3SP\mh\server.py", line 273, in _worker_target
    handler.on_packet(packet)
  File "C:\Users\sepal\Projects\MHTri Project\MH3SP\mh\pat.py", line 2493, in on_packet
    self.dispatch(packet_id, data, seq)
  File "C:\Users\sepal\Projects\MHTri Project\MH3SP\mh\pat.py", line 2516, in dispatch
    return handler(packet_id, data, seq)
  File "C:\Users\sepal\Projects\MHTri Project\MH3SP\mh\pat.py", line 284, in recvReqAuthenticationToken
    self.sendAnsAuthenticationToken(nas_token, seq)
  File "C:\Users\sepal\Projects\MHTri Project\MH3SP\mh\pat.py", line 295, in sendAnsAuthenticationToken
    self.send_packet(PatID4.AnsAuthenticationToken, b'', seq)
  File "C:\Users\sepal\Projects\MHTri Project\MH3SP\mh\pat.py", line 2496, in send_packet
    super(PatRequestHandler, self).send_packet(packet_id, data, seq)
  File "C:\Users\sepal\Projects\MHTri Project\MH3SP\mh\server.py", line 111, in send_packet
    self.wfile.write(data)
  File "C:\Python39\lib\socket.py", line 722, in write
    return self._sock.send(b)
  File "C:\Python39\lib\ssl.py", line 1173, in send
    return self._sslobj.write(data)
ssl.SSLEOFError: EOF occurred in violation of protocol (_ssl.c:2477)
```

Still working fine with python2. I'll investigate and do some more testing.

PR #52, opened 2022-08-21
Properly send the weapon type in the CircleInfo

InusualZ:
Before (from the Test Server):
image

After:
image

### DIFF
--- a/mh/session.py
+++ b/mh/session.py
@@ -315,10 +315,9 @@ class Session(object):
     def get_optional_fields(self):
         """LayerUserInfo's optional fields."""
         location = int(self.is_in_quest())  # City - 0, Quest - 1
-        hunter_rank, = struct.unpack_from(">H", self.hunter_info.data, 0)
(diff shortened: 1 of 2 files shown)


mh/session.py
```diff
@@ -315,10 +315,9 @@ def get_layer_host_data(self):
     def get_optional_fields(self):
         """LayerUserInfo's optional fields."""
         location = int(self.is_in_quest())  # City - 0, Quest - 1
-        hunter_rank, = struct.unpack_from(">H", self.hunter_info.data, 0)
-        weapon_icon, = struct.unpack_from(">B", self.hunter_info.data, 0x10)
-        weapon_icon -= 7  # Skip armor pieces and start at index zero
```
sepalani, line 320: Won't removing this `-7` part break the LayerUserInfo?

InusualZ: (I tested it)

No, because this is no the weapon type ID from the first equipment slot. This is another field in the ‘hunter_info’ that’s send specifically for this purpose. If we were to use the first equipment’s weapon type ID we would have to calculate bowguns (light/medium/heavy)

PR #53, opened 2022-08-21
Re-add timeout workaround

InusualZ:
I was wrong, it still needed :(

Let's not have a broken `dev` branch.


sepalani:
Haven't investigated it yet but the revert shouldn't be harmful anyway. LGTM.

PR #55, opened 2022-08-26
Add legacy SSL ciphers support

sepalani:
This PR fixes issues related to SSL security being too weak. It targets the master branch as some users are concerned especially when running the server on recent OS and Linux distributions. I chose to disable the option by default since I don't think it's a good idea to bypass existing security policies and people doing so should be aware of that.

Ready to reviewed & merged.


InusualZ (approved):
LGTM

PR #56, opened 2022-08-29
Added code for disconnecting from layers

kbarkevich:
… through means other than the city entrance

This now includes code for city owners to transfer ownership upon disconnection.

### DIFF
--- a/mh/pat.py
+++ b/mh/pat.py
@@ -600,18 +600,21 @@ class PatRequestHandler(SocketServer.StreamRequestHandler, object):
         ID: 60100100
         JP: 切断要求
         TR: Disconnection request
-        """
-        login_type, = struct.unpack(">B", data)
-        self.sendAnsShut(login_type, seq)
 
-    def sendAnsShut(self, login_type, seq):
+        # 1: full logout(?)
+        # 2: logout to different server(?)
+        """
+        shutdown_type, = struct.unpack(">B", data)
+        self.sendAnsShut(shutdown_type, seq)
+
+    def sendAnsShut(self, shutdown_type, seq):
         """AnsShut packet.
 
         ID: 60100200
         JP: 切断返答
         TR: Disconnection response
         """
-        data = struct.pack(">B", login_type)
+        data = struct.pack(">B", shutdown_type)
         self.send_packet(PatID4.AnsShut, data, seq)
         self.finish()
 
@@ -822,7 +825,7 @@ class PatRequestHandler(SocketServer.StreamRequestHandler, object):
         self.server.debug("UserObject: {}, {}, {!r}".format(
             is_slot_empty, slot_index, user_obj
         ))
-        hunter_name = ""
+        hunter_name = None
         if hasattr(user_obj, "hunter_name"):
             hunter_name = pati.unpack_string(user_obj.hunter_name)
         self.session.use_user(slot_index, hunter_name)
@@ -1004,6 +1007,7 @@ class PatRequestHandler(SocketServer.StreamRequestHandler, object):
         JP: レイヤ終了要求
         TR: Layer end request
         """
+        self.notify_layer_departure()
         self.sendAnsLayerEnd(seq)
 
     def sendAnsLayerEnd(self, seq):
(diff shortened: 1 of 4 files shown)


mh/session.py
```diff
@@ -303,6 +327,17 @@ def get_layer_players(self):
         else:
             assert False, "Can't find layer"
 
+    def attempt_leave_all_layers(self, shutdown_type):
+        assert 1 <= shutdown_type <= 2, "Invalid shutdown type"
+        if shutdown_type != 1:
+            return  # only shutdown_type 1 requires removal from layers
```
sepalani, line 333: Please add a comment to explain the differences between `layer_end` and this function including which particular issues/cases it addresses. I want to avoid repeating the same mistakes ><'.

Could you please move the `assert` and `shutdown_type` check and perform them in `recvReqShut` instead. That will get rid of the parameter.

sepalani (changes requested):
Could you please rebase your PR on top of `dev` again, it seems a force-push included unrelated commits.

InusualZ:
Okay, So, like I say on discord I have taken over this PR. I have rebased and pushed the new approach. This new approach consist of **always** doing the session cleanup. Unless, a flag (`Session::request_reconnection`) is set, but even if that flag is set we would do the session cleanup as usual, it just that we wouldn't erase the `session` from the `db` and we wouldn't detach the session from the `server` (`Valor 1`, `Valor 2`, etc...)

We only set that flag (`Session::request_reconnection`) to `True` in two places:
- When a totally new (No `online_support_code`) [client arrive](https://github.com/sepalani/MH3SP/blob/9fb602683273e9c916a8928296206bb3e675cf9c/mh/pat.py#L289) to the (OPN) server, because we need the session to be kept alive for the total duration of the login process (From `OPN`. To `FMP`) due to the client only saving the `online_support_code` only when it reaches the `FMP` server.
- When the client [request information](https://github.com/sepalani/MH3SP/blob/9fb602683273e9c916a8928296206bb3e675cf9c/mh/pat.py#L1049) about the `FMP` server that he would be connecting to. This because we need to keep track of the `FMP` server that the client wants to connect.

This flag (`Session::request_reconnection`) would always be reset once the [client connect](https://github.com/sepalani/MH3SP/blob/9fb602683273e9c916a8928296206bb3e675cf9c/mh/pat.py#L289) to the desired FMP Server (At this point the condition would always return `False` because the client finished the login process)

This implementation would allows us to leak less `Session` objects.

mh/pat.py
```diff
@@ -286,6 +286,9 @@ def sendNtcLogin(self, server_status, connection_data, seq):
         """
         data = struct.pack(">B", server_status)
         self.session = self.session.get(connection_data)
+        self.session.request_reconnection = \
+            not ("pat_ticket" in connection_data or
+                 "online_support_code" in connection_data)
```
sepalani, line 291: Do you think doing this part in `session.get` is possible?

mh/pat.py
```diff
@@ -1039,6 +1046,7 @@ def recvReqFmpInfo(self, packet_id, data, seq):
             self.sendAnsFmpInfo(fmp_data, fields, seq)
         elif packet_id == PatID4.ReqFmpInfo2:
             self.sendAnsFmpInfo2(fmp_data, fields, seq)
+        self.session.request_reconnection = True
```
sepalani, line 1049: Maybe add a comment to explain why this is needed (i.e. to preserve the session at the server selection screen).

mh/session.py
```diff
@@ -223,7 +235,7 @@ def find_users(self, capcom_id, hunter_name, first_index, count):
         return users[start:start+count]
 
     def leave_server(self):
-        DB.leave_server(self)
+        DB.leave_server(self, self, self.local_info["server_id"])
```
sepalani, line 238: Is `self, self,` a typo?

sepalani (changes requested):
I really like the new approach. Could you please squash all the commits into a single one?

sepalani (approved):
Code LGTM, untested.

PR #57, opened 2022-09-01
City creation and joining bug

kbarkevich:
Added code to prevent the creation of invalid cities.
This included cities that already exist, or are already reserved by another user in the city creation menu.

This would fix, among others, the bug pictured below.

image

This does NOT implement city list refreshing. It simply removes the more severe consequences of the city list not refreshing, such as server/client crashes and layer misplacement.

### DIFF
--- a/mh/database.py
+++ b/mh/database.py
@@ -20,6 +20,7 @@
 """
 
 import random
+import time
 
 CHARSET = "0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZ"
 
@@ -29,6 +30,8 @@ MEDIA_VERSIONS = {
     "1001301045": "RMHE08",
     "1002121503": "RMHP08",
 }
+
+RESERVE_DC_TIMEOUT = 40.0
 
 
 def new_random_str(length=6):
@@ -202,6 +205,7 @@ class City(object):
         self.players = Players(4)
         self.optional_fields = []
         self.leader = None
+        self.reserved = None
         self.circles = [
             # One circle per player
             Circle(self) for _ in range(self.get_capacity())
@@ -244,6 +248,12 @@ class City(object):
             if circle.leader == leader_session:
                 return circle, index
         return None, None
+
+    def reserve(self, reserve):
+        if reserve:
+            self.reserved = time.time()
+        else:
+            self.reserved = None
 
 
 class Gate(object):
@@ -496,6 +506,14 @@ class TempDatabase(object):
         cities = self.get_cities(server_id, gate_id)
         assert 0 < index <= len(cities), "Invalid city index"
         return cities[index - 1]
+
+    def reserve_city(self, server_id, gate_id, index, reserve):
+        city = self.get_city(server_id, gate_id, index)
+        reserved_time = city.reserved
+        if reserve and reserved_time and time.time()-reserved_time < RESERVE_DC_TIMEOUT:
+            return False
+        city.reserve(reserve)
+        return True
 
     def get_all_users(self, server_id, gate_id, city_id):
         """Search for users in layers and its children.
--- a/mh/pat.py
+++ b/mh/pat.py
@@ -2198,6 +2198,12 @@ class PatRequestHandler(SocketServer.StreamRequestHandler, object):
         TR: Layer creation response
         """
         data = struct.pack(">H", number)
+        if not self.session.is_city_empty(number):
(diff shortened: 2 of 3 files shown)


mh/session.py
```diff
@@ -240,6 +240,15 @@ def get_cities(self):
...
 
+    def is_city_empty(self, city_id):
+        return self.get_gate().cities[city_id-1].get_state() == db.LayerState.EMPTY
+
+    def reserve_city(self, city_id, reserve):
+        self.get_gate().cities[city_id-1].reserve(reserve)
+
+    def is_city_reserved(self, city_id):
+        return self.get_gate().cities[city_id-1].reserved
+
```
sepalani, line 251: These should have a proper implementation in the database. You might to think `session` as cache that will interact with the database only when necessary. Your current approach is only working because we're using a temporary database that's directly accessible.

```python
    def is_city_empty(self, city_id):
        return DB.get_city(self.local_info["server_id"], self.local_info["gate_id"], city_id).get_state() == db.LayerState.EMPTY

    def reserve_city(self, city_id, reserve):
        return DB.reserve_city(self.local_info["server_id"], self.local_info["gate_id"], city_id, reserve)

    def is_city_reserved(self, city_id):
        return DB.is_city_reserved(self.local_info["server_id"], self.local_info["gate_id"], city_id)
```

I also don't want to expose implementation detail in the `session` module, for instance, `id` starting at 0 internally whereas in the game they usually start at 1.

mh/pat.py
```diff
@@ -2198,6 +2198,13 @@ def sendAnsLayerCreateHead(self, number, seq):
         TR: Layer creation response
         """
         data = struct.pack(">H", number)
+        if not self.session.is_city_empty(number):
+            self.sendAnsAlert(PatID4.AnsLayerCreateHead, "<LF=8><BODY><CENTER>City already exists.<END>", seq)
+            return
+        if self.session.is_city_reserved(number):
+            self.sendAnsAlert(PatID4.AnsLayerCreateHead, "<LF=8><BODY><CENTER>City reserved.<END>", seq)
+            return
+        self.session.reserve_city(number, True)
```
sepalani, line 2207: Can't `reserve_city` returns a boolean to show when it's successful? I doubt there is another place where we'd need such check. It will also save a call to the database.

```python
if not self.session.reserve_city(number, True):
    self.sendAnsAlert(PatID4.AnsLayerCreateHead, "<LF=8><BODY><CENTER>City already reserved.<END>", seq)
    return
```

sepalani (changes requested):
The logic looks sane to me, otherwise.

mh/pat.py
kbarkevich: Addressed in latest commit

mh/session.py
kbarkevich: Addressed in latest commit

kbarkevich:
It has also occurred to me that we might need some sort of timeout mechanism for city reservations. If somebody disconnects from the server unexpectedly while having a reserved city, it could stay reserved forever. Maybe a hard threshold of 40 seconds (as opposed to the client-side 30 seconds) that, if the city has been reserved for longer than that, it can be reserved again.

Either that or we would need to associate reservations with players so we can keep track of them and cancel them should an unexpected disconnect occur.

InusualZ (approved):

mh/database.py
```diff
@@ -497,6 +507,14 @@ def get_city(self, server_id, gate_id, index):
         assert 0 < index <= len(cities), "Invalid city index"
         return cities[index - 1]
 
+    def reserve_city(self, server_id, gate_id, index, reserve):
+        reserved_status = self.get_city(server_id, gate_id, index).reserved
+        if reserve is True and reserved_status is not None and\
+            time.time()-reserved_status < RESERVE_DC_TIMEOUT:
+            return False
+        self.get_city(server_id, gate_id, index).reserve(reserve)
+        return True
```
sepalani, line 516: When possible, let's avoid calling functions more than necessary. The following should be equivalent:
```python
def reserve_city(self, server_id, gate_id, index, reserve):
    city = self.get_city(server_id, gate_id, index)
    reserve_time = city.reserved
    if reserve and reserve_time and time.time()-reserve_time < RESERVE_DC_TIMEOUT:
        return False
    city.reserve(reserve)
    return True
```

kbarkevich: Fixed

sepalani (commented):
Would LGTM when all these commits are squashed into a single one with a proper commit message.

PR #63, opened 2022-09-11
Allow all supported SSL protocols

sepalani:
Not sure why enforcing TLSv1 work in some cases and don't in others for some people. Switching this PR as draft while I make sure it doesn't cause a regression from all my testing devices (i.e. Dolphin/real Wii and VPS/local server).

### DIFF
--- a/other/utils.py
+++ b/other/utils.py
@@ -253,7 +253,10 @@ def create_server(server_class, server_handler,
     if use_ssl:
         import ssl
 
-        context = ssl.SSLContext(ssl.PROTOCOL_TLSv1)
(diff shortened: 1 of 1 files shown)


sepalani:
For some reasons I was unable to make any protocol to work with Python 2.7 except PROTOCOL_SSLv23 whereas Python 3 was working as intended.

Rather than trying to find what was the issue with Python 2.7 I rolled back to the previous protocol which was working except for Python 3 w/ recent versions of OpenSSL and I addressed the Python 3 issue instead.

InusualZ (approved):

PR #64, opened 2022-09-20
Added time passing and 14-day cycle for Jhen event

kbarkevich:
Implemented serverside timers for day cycle ticks (counts backwards from epoch time) as well as server time (counts forwards from epoch time).

Also implemented cyclical Jhen Mohran event, which is based on server time and repeats every 14 days. This cycle starts with 1 day of fog followed by 2 days of sandstorm, and then 11 days of calm before repeating.

### DIFF
--- a/mh/constants.py
+++ b/mh/constants.py
@@ -21,19 +21,14 @@
 
 import struct
 from other.utils import pad
+from mh.time_utils import current_tick, TICKS_PER_CYCLE, get_jhen_event_times
 from quest_utils import QUEST_EVENT_JUMP_FOUR_JAGGI, QUEST_EVENT_BLOOD_SPORT,\
     QUEST_EVENT_MERCY_MISSION, QUEST_EVENT_THE_PHANTOM_URAGAAN, QUEST_EVENT_WORLD_EATER,\
     QUEST_EVENT_WHERE_GODS_FEAR_TO_TREAD, QUEST_EVENT_GREEN_EGGS
 
 
-def make_binary_type_time_events(state=0):
-    # TODO: Use the server in-game time and handle this time array properly
-    if state == 1:
-        return struct.pack(">iii", 0, -1, 0)  # Fog
-    elif state == 2:
-        return struct.pack(">iii", 0, 0, -1)  # Kujira
-    else:
-        return struct.pack(">iii", 0, 0, 0)
+def make_binary_type_time_events():
+    return struct.pack(">III", *get_jhen_event_times())
 
 
 def make_binary_server_type_list(is_jap=False):
@@ -59,8 +54,9 @@ def make_binary_server_type_list(is_jap=False):
                 desc[i:i+1] = b' ' * padding
         data += pad(desc, 112 if is_jap else 168)
 
-    # TODO: Figure out what it is
-    data += b"\0" * 12
+    data += struct.pack(">I", 0)  # unk
+    data += struct.pack(">I", current_tick())  # Current time tick, counts backwards
+    data += struct.pack(">I", TICKS_PER_CYCLE)  # Max Tick Per Cycle, if 0 game defaults to 3000
 
     # Handle city seekings (x32)
     SEEKINGS = [
@@ -259,11 +255,11 @@ OTHER_HUNTER_NAME = b"Drakea"
 PAT_BINARIES = {
     0x01: {
         "version": 1,
-        "content": make_binary_server_type_list(is_jap=IS_JAP)
+        "content": lambda: make_binary_server_type_list(is_jap=IS_JAP)
     },
     0x02: {
         "version": 1,
-        "content": make_binary_type_time_events(state=TIME_STATE)
+        "content": make_binary_type_time_events
     },
     0x03: {
         "version": 1,
@@ -321,11 +317,11 @@ PAT_BINARIES = {
     },
     0x10: {  # French
         "version": 1,
-        "content": make_binary_server_type_list()
+        "content": make_binary_server_type_list
     },
     0x11: {  # French
         "version": 1,
-        "content": make_binary_type_time_events(state=TIME_STATE)
+        "content": make_binary_type_time_events
     },
     0x12: {  # French
         "version": 1,
@@ -381,11 +377,11 @@ PAT_BINARIES = {
     },
     0x1f: {  # German
         "version": 1,
-        "content": make_binary_server_type_list()
+        "content": make_binary_server_type_list
     },
     0x20: {  # German
         "version": 1,
-        "content": make_binary_type_time_events(state=TIME_STATE)
+        "content": make_binary_type_time_events
     },
     0x21: {  # German
         "version": 1,
@@ -441,11 +437,11 @@ PAT_BINARIES = {
     },
     0x2e: {  # Italian
         "version": 1,
-        "content": make_binary_server_type_list()
+        "content": make_binary_server_type_list
     },
     0x2f: {  # Italian
         "version": 1,
-        "content": make_binary_type_time_events(state=TIME_STATE)
+        "content": make_binary_type_time_events
     },
     0x30: {  # Italian
         "version": 1,
@@ -501,11 +497,11 @@ PAT_BINARIES = {
     },
     0x3d: {  # Spanish
         "version": 1,
-        "content": make_binary_server_type_list()
+        "content": make_binary_server_type_list
     },
     0x3e: {  # Spanish
         "version": 1,
-        "content": make_binary_type_time_events(state=TIME_STATE)
+        "content": make_binary_type_time_events
     },
     0x3f: {  # Spanish
         "version": 1,
--- a/mh/pat.py
+++ b/mh/pat.py
@@ -806,9 +806,9 @@ class PatRequestHandler(SocketServer.StreamRequestHandler, object):
         JP: サーバ時刻返答
         TR: Server time response
         """
-        dummy_1 = 0  # in-game time?
-        current_time = time_utils.datetime_to_int(datetime.now())
-        data = struct.pack(">II", dummy_1, current_time)
+        server_time = time_utils.current_server_time()  # Counter that ticks up once per second, Epoch works
+        current_time = time_utils.current_server_time()  # Always Epoch, used for Wii Shop subscription ticket
+        data = struct.pack(">II", server_time, current_time)
         self.send_packet(PatID4.AnsServerTime, data, seq)
 
     def recvReqUserObject(self, packet_id, data, seq):
@@ -1095,7 +1095,10 @@ class PatRequestHandler(SocketServer.StreamRequestHandler, object):
          - Spanish: 0x41, 0x3d, 0x41, 0x3e, 0x3f, 0x40
         """
         binary = PAT_BINARIES[binary_type]
-        data = struct.pack(">II", binary["version"], len(binary["content"]))
+        content = binary["content"]
+        if callable(content):
+            content = content()
+        data = struct.pack(">II", binary["version"], len(content))
         self.send_packet(PatID4.AnsBinaryHead, data, seq)
 
     def recvReqBinaryData(self, packet_id, data, seq):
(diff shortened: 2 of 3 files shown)


mh/constants.py
```diff
@@ -257,13 +263,13 @@ def slot(item, qty):
 
 # Dummy PAT_BINARY
 PAT_BINARIES = {
-    0x01: {
+    0x01: lambda: {
         "version": 1,
         "content": make_binary_server_type_list(is_jap=IS_JAP)
     },
```
sepalani, line 269: I don't think converting the whole object to a lambda returning a dict/object is a nice idea, especially since the `version` field is a constant. Could you please use a lambda or function pointer in the `content` field instead?

mh/pat.py
```diff
@@ -806,9 +806,9 @@ def sendAnsServerTime(self, converted_country_code, seq):
         JP: サーバ時刻返答
         TR: Server time response
         """
-        dummy_1 = 0  # in-game time?
-        current_time = time_utils.datetime_to_int(datetime.now())
-        data = struct.pack(">II", dummy_1, current_time)
+        server_time = time_utils.current_server_time()  # Epoch time, counts forward at 1 per second
```
sepalani, line 809: AFAIK, we don't have such evidence that the "server_time" is based on epoch so I don't think such comment is appropriate. However, we do know that the `current_time` used for Wii Shop's ticket is using epoch.

image

mh/time_utils.py
```diff
@@ -20,9 +20,44 @@
...
+    return next_day*24*60*60
+
+def get_fog_start(time=current_server_time()):
+    return int(get_jhen_cycle_time(time) + FOG_START*24*60*60)
+
+def get_sandstorm_start(time=current_server_time()):
+    return int(get_jhen_cycle_time(time) + JHEN_START*24*60*60)
+
+def get_sandstorm_end(time=current_server_time()):
+    return int(get_jhen_cycle_time(time) + JHEN_END*24*60*60)
```
sepalani, line 63: - Can't these be a simple function returning a tuple instead? You're calling the same function multiple times.
 - `time` as a variable name should be avoided as it's shadowing the `time` module.
 - `24 * 60 * 60` should be stored to a variable or a constant with a meaningful name.
 - IIRC, you don't need to import `floor` in your case and can use the `//` operator: `time // (24*60*60)`
 - Can't a modulo be used instead of the `if (...) >= JHEN_END: (...) else:`?

mh/time_utils.py
```diff
@@ -20,9 +20,44 @@
...
     return int((dt - EPOCH).total_seconds())
+
+JHEN_EVENT_OFFSET = 14  # Cycle of 14 days
+FOG_START = 0  
+JHEN_START = 1  # 1 day of fog
+JHEN_END = 3  # Followed by 2 days of sandstorm
+
+
+def current_tick():
+    return 4294967295 - datetime_to_int(datetime.datetime.now())
```
sepalani, line 40: Please use the hex notation (0xffffffff) to make it clearer and add some docstring to explain your rational. I have many doubts regarding it to be the right approach. Though, I'll let it slide for the time being since we don't have anything better.

mh/constants.py
```diff
@@ -21,19 +21,24 @@
...
-        return struct.pack(">iii", 0, 0, 0)
+def make_binary_type_time_events():
+    """
+    If the first int is less than the gametime at server login:
+      If the second int is greater than the gametime at server login, fog
+      Otherwise, if the third int is greater than the gametime at server login, sandstorm
+    FIRST INT: Start of fog (epoch seconds)
+    SECOND INT: Start of Jhen event (epoch seconds)
+    THIRD INT: End of Jhen event (epoch seconds)
+    """
```
sepalani, line 39: This docstring should be moved to the util function in `mh.time_utils`, IMHO. So it would look like something like this where `get_time_events` is the function returning the tuple of events:
```python
def make_binary_type_time_events():
   return struct.pack(">III", *get_time_events())
```

sepalani (changes requested):

InusualZ:
I concur with all of @sepalani points.

mh/constants.py
kbarkevich: Fixed this, instead the content is now (usually) just the function definition so it can be called without a lambda. The exception is the first one, where the is_jap parameter is specified.

mh/pat.py
kbarkevich: Changed the comment

mh/time_utils.py
kbarkevich: * Turned into a single function returning a tuple
* No longer shadowing `time` module
* Stored `24 * 60 * 60` to a constant
* Removed `floor` import and switched to `//`
* I'm not sure what you mean by this one, I'm already using a modulus in the `if` statement to determine if the day in the event cycle is passed the end of the jhen event

mh/time_utils.py
kbarkevich: Switched to hex notation and added docstring explaining the current rationale

mh/constants.py
kbarkevich: Addressed this while addressing a previous comment. Also moved the docstring.

mh/time_utils.py
```diff
@@ -21,8 +21,68 @@
...
 
+
 EPOCH = datetime.datetime(1970, 1, 1)
+TICKS_PER_CYCLE = 5400  # 1.5 hours per cycle, 45 mins per day
 
 
 def datetime_to_int(dt):
     return int((dt - EPOCH).total_seconds())
+
+SECONDS_IN_DAY = 24*60*60
```
sepalani, line 32: `SECONDS_PER_DAY` would be clearer.

mh/pat.py
```diff
@@ -1095,6 +1095,8 @@ def sendAnsBinaryHead(self, binary_type, seq):
          - Spanish: 0x41, 0x3d, 0x41, 0x3e, 0x3f, 0x40
         """
         binary = PAT_BINARIES[binary_type]
+        if callable(binary["content"]):
+            binary["content"] = binary["content"]()
```
sepalani, line 1099: Since dictionaries are using references, I think you're overriding the function with its value so it won't be called ever again.
```python
content = binary["content"]
if callable(content):
    content = content()
```

mh/pat.py
```diff
@@ -1109,6 +1111,8 @@ def recvReqBinaryData(self, packet_id, data, seq):
         """
         binary_type, version, offset, size = struct.unpack(">BIII", data)
         binary = PAT_BINARIES[binary_type]
+        if callable(binary["content"]):
+            binary["content"] = binary["content"]()
```
sepalani, line 1115: Ditto.

mh/time_utils.py
```diff
@@ -21,8 +21,68 @@
...
+    server is shut down.
+    """
+    return 0xFFFFFFFF - datetime_to_int(datetime.datetime.now())
+
+
+def current_server_time():
+    return datetime_to_int(datetime.datetime.now())
+
+
+def get_jhen_cycle_time(curr_time=current_server_time()):
```
sepalani, line 67: IIRC, `curr_time=current_server_time()` will only be evaluated once, during the definition of the function. You should get rid of this default parameter and either call it explicitly in the function or pass it as parameter.

sepalani (changes requested):

mh/time_utils.py
kbarkevich: Fixed

mh/pat.py
kbarkevich: You're right, I've changed it to a way that does not override the value in the dictionary.

image

mh/pat.py
kbarkevich: Fixed as per above

mh/time_utils.py
kbarkevich: Fixed behavior by changing the default parameter to `None`

mh/time_utils.py
```diff
@@ -21,8 +21,72 @@
...
+    server is shut down.
+    """
+    return 0xFFFFFFFF - datetime_to_int(datetime.datetime.now())
+
+
+def current_server_time():
+    return datetime_to_int(datetime.datetime.now())
+
+
+def get_jhen_cycle_time(curr_time=None):
```
sepalani, line 67: This function is only used here and doesn't make much sense on its own. You might want to move its content into `get_jhen_event_times` directly.

mh/time_utils.py
```diff
@@ -21,8 +21,72 @@
...
+        curr_time = current_server_time()
+    current_server_day = int(curr_time//SECONDS_PER_DAY)
+    if (current_server_day % JHEN_EVENT_OFFSET) >= JHEN_END:
+        next_day = current_server_day + (JHEN_EVENT_OFFSET - (current_server_day % JHEN_EVENT_OFFSET))
+    else:
+        next_day = current_server_day - (current_server_day % JHEN_EVENT_OFFSET)
+    
+    return next_day*SECONDS_PER_DAY
+
+def get_jhen_event_times(curr_time=None):
```
sepalani, line 78: This parameter is never passed so you might remove it and set it in the function directly.

mh/time_utils.py
```diff
@@ -21,8 +21,72 @@
...
+def get_jhen_cycle_time(curr_time=None):
+    if curr_time is None:
+        curr_time = current_server_time()
+    current_server_day = int(curr_time//SECONDS_PER_DAY)
+    if (current_server_day % JHEN_EVENT_OFFSET) >= JHEN_END:
+        next_day = current_server_day + (JHEN_EVENT_OFFSET - (current_server_day % JHEN_EVENT_OFFSET))
+    else:
+        next_day = current_server_day - (current_server_day % JHEN_EVENT_OFFSET)
+    
+    return next_day*SECONDS_PER_DAY
```
sepalani, line 76: By having some intermediate variables, you'll make your intent clearer:
```python
def get_jhen_event_times():
    current_day = int(current_server_time() // SECONDS_PER_DAY)
    day_offset = current_day % JHEN_EVENT_OFFSET
    if day_offset < JHEN_END:
        period_start = (current_day - day_offset) * SECONDS_PER_DAY
    else:  # Upcoming period
        period_start = (current_day - day_offset + JHEN_EVENT_OFFSET) * SECONDS_PER_DAY
    return (
        period_start + FOG_START*SECONDS_PER_DAY,
        period_start + JHEN_START*SECONDS_PER_DAY,
        period_start + JHEN_END*SECONDS_PER_DAY
    )
```

sepalani (changes requested):
Beside that, the code LGTM, I haven't tested it yet.

mh/time_utils.py
kbarkevich: I think there's value in the readability that comes with separating this logic out into another function, and if we want to base other event quest timings in the cycle on this one "starting" time we're going to want it separated out. Right now it's just the Jhen quest that we're timing based on this cycle, but in the future we're going to need some sort of rotation, and I think having this separated out will help with that when we get there.

kbarkevich: Actually I take that back, this is something that should be worried about once we actually implement the rest of the event quest cycle system.

mh/time_utils.py
kbarkevich: Done

mh/time_utils.py
kbarkevich: Done

mh/pat.py
```diff
@@ -806,9 +806,9 @@ def sendAnsServerTime(self, converted_country_code, seq):
         JP: サーバ時刻返答
         TR: Server time response
         """
-        dummy_1 = 0  # in-game time?
-        current_time = time_utils.datetime_to_int(datetime.now())
-        data = struct.pack(">II", dummy_1, current_time)
+        server_time = time_utils.current_server_time()
+        current_time = time_utils.current_server_time()  # Epoch time
```
InusualZ, line 810: Please add a comment explaining why this values are the same but means different things.

InusualZ (changes requested):
Just one comment, everything else LGTM

mh/pat.py
kbarkevich: Done

mh/constants.py
kbarkevich: After thinking about this some more, I think having the entire dict be a `lambda`-generated dict will end up being the way to go. This is because, as event quests cycle through, the version is going to change. Version 1 would be for the first set of event quests, version 2 would be for the second set of event quests, etc.

This doesn't pertain to this particular PR so I'll leave it as is, but we're probably going to need to revert this or otherwise dynamically generate the `version` as well as the `content` once a proper event quest cycle is created.

InusualZ (approved):

sepalani (approved):
Code LGTM and tested. Working on both western versions of the game. Not working on the Japanese version, the event quest list isn't populated. I'm going to merge this PR anyway and open an issue for it.

PR #75, opened 2022-11-09
Added timer-based & console-based announcement support

kbarkevich:
This pull requests implements a basic server-wide announcement system, similar to that which Capcom used to display their "Capcom says:" messages.

* One basic "thanks for playing" message is scheduled to dispatch every 30 minutes
* Added `broadcast_message` function, callable from the `--interactive` console, to send any message to all online users


fmp_server.py
```diff
@@ -26,9 +26,61 @@
...
+    curr_broadcast = mod_words[0]
+    for word in mod_words[1:]:
+        if len(curr_broadcast) + 1 + len(word) <= 30:
+            curr_broadcast += " "
+            curr_broadcast += word
+        else:
+            broadcast_segments.append(curr_broadcast)
+            curr_broadcast = word
+    if len(curr_broadcast) > 0:
+        broadcast_segments.append(curr_broadcast)
```
sepalani, line 54: - Isn't there a simpler way to perform this? It looks like a lot of code for a simple task.
- Didn't Capcom used newlines instead of creating a message for each line?

fmp_server.py
```diff
@@ -26,9 +26,61 @@
...
+    
+    broadcasts = []
+    for broadcast_text in broadcast_segments:
+        data = struct.pack(">B", 0x01)
+        info = pati.MessageInfo()
+        info.text_color = pati.Long(0x00fffff0)
+        info.sender_id = pati.String("000000")
+        info.sender_name = pati.String("[MH3SP]")
```
sepalani, line 62: These should probably be optional parameters too.

fmp_server.py
```diff
@@ -26,9 +26,61 @@
...
+        data += info.pack()
+        data += pati.lp2_string(broadcast_text)
+        print("BROADCAST: " + broadcast_text)
```
sepalani, line 65: Can't we log this message properly, somewhere, rather than using raw prints?

master_server.py
```diff
@@ -47,13 +54,21 @@ def main(args):
     """Master server main function."""
     servers, has_ui = create_servers(silent=args.silent,
                                      debug_mode=args.debug_mode)
+    broadcast_queues = {"FmpServer": queue.Queue(), "OpnServer": queue.Queue(),
+                        "LmpServer": queue.Queue(), "RfpServer": queue.Queue()}
     threads = [
-        threading.Thread(target=server.serve_forever)
+        threading.Thread(target=server.serve_forever, name=server.__class__.__name__,
+            args=(broadcast_queues[server.__class__.__name__], ))
```
sepalani, line 61: If you plan to use `server.__class__.__name__` then you should also use it in `broadcast_queues` rather than hardcoding the (dict's key) string.

master_server.py
```diff
@@ -47,13 +54,21 @@ def main(args):
...
         for server in servers
     ]
     for thread in threads:
         thread.start()
 
+    def broadcast_message(text):
+        broadcast = FMP.construct_broadcast(text)
+        for broadcast_segment in broadcast:
+            broadcast_queues["FmpServer"].put(broadcast_segment, block=True)
+
```
sepalani, line 71: Maybe you should add a comment stating that it's for easier use in interactive mode. Moreover, you shouldn't use `"FmpServer"` string and prefer using `.__class__.__name__` when possible.

mh/server.py
```diff
@@ -266,13 +277,20 @@ def _worker_target(self, work_queue):
...
-
-            try:
-                handler.on_packet(packet)
-            except Exception as e:
-                handler.on_exception(e)
+            
+            if event == selectors.EVENT_WRITE:
+                try:
+                    if handler.session.layer > 0:  # in-game only
+                        handler.send_packet(packet[0], packet[1], packet[2])
```
sepalani, line 284: If packet is a tuple of 3 elements, use the `*` operator instead: `*packet`.

sepalani (changes requested):
I'm going to test it against the Japanese version, asap. This version tended to have different char limit for message boxes.

fmp_server.py
kbarkevich: It has to trim down lengthy words to segments that are at most the maximum character length, then do the math word-by-word to see if each one fits onto the current line when including a space. Those are 2 distinct processes, thus the 2 different for loops.

I DID completely forget to use the constant `MAX_CHAR_LENGTH` I defined though. So, oops.

kbarkevich: Also, I don't think Capcom did? The messages always showed up as distinct lines, each with their own sender name.

fmp_server.py
kbarkevich: By the time it gets to the FmpServer, it's already bundled up into a binary. This function doesn't have access to any loggers since it's not local to any of the `PatServer` classes.

fmp_server.py
kbarkevich: Done

master_server.py
kbarkevich: Done

master_server.py
kbarkevich: Done

mh/server.py
kbarkevich: Done

fmp_server.py
sepalani: > Also, I don't think Capcom did? The messages always showed up as distinct lines, each with their own sender name.

That seems correct, they are sending 4 messages as part of their announcement as seen on these videos:
 - https://youtu.be/ZTwdF9e9Kk0?t=2943
 - https://youtu.be/K5dX4KhUFUM?t=1570

image
image

fmp_server.py
```diff
@@ -26,9 +26,62 @@
...
+    curr_broadcast = mod_words[0]
+    for word in mod_words[1:]:
+        if len(curr_broadcast) + 1 + len(word) <= MAX_CHAR_LENGTH:
+            curr_broadcast += " "
+            curr_broadcast += word
+        else:
+            broadcast_segments.append(curr_broadcast)
+            curr_broadcast = word
+    if len(curr_broadcast) > 0:
+        broadcast_segments.append(curr_broadcast)
```
sepalani, line 55: Can't the module `textwrap` and `textwrap.wrap` be used instead?

fmp_server.py
sepalani: Is this print necessary since the server could log any packets including broadcast ones?

master_server.py
```diff
@@ -47,13 +54,21 @@ def main(args):
...
         for server in servers
     ]
     for thread in threads:
         thread.start()
 
+    def broadcast_message(text):
+        # Function for easier broadcasting in interactive mode.
+        broadcast = FMP.construct_broadcast(text)
+        for broadcast_segment in broadcast:
+            broadcast_queues[FMP.BASE.cls.__name__].put(broadcast_segment, block=True)
```
sepalani, line 70: Can't we iterate over the servers and call their `broadcast_message` method rather than "hijacking" their `broadcast_queues`?

mh/server.py
```diff
@@ -206,19 +207,24 @@ def fileno(self):
         """
         return self.socket.fileno()
 
-    def serve_forever(self):
+    def serve_forever(self, broadcast_queue):
+        self.broadcast_queue = broadcast_queue
```
sepalani, line 211: Passing the `broadcast_queue` as a parameter seems overkill. Can't it be created in the server's constructor method?

mh/server.py
```diff
@@ -206,19 +207,24 @@ def fileno(self):
...
         self.__is_shut_down.clear()
         try:
             with self.selector as selector:
                 selector.register(self, selectors.EVENT_READ)
 
+                event_watch = Timer()
```
sepalani, line 217: It will do for the time being but a clock should be used. Having a single timer prevent handling multiple timed events.

sepalani (changes requested):
A very quick review.

PR #78, opened 2022-11-11
Added dynamically updating greeter text

kbarkevich:
Greeter text changes based on sandstorm status. This can be further updated to support automatically-changing event quests (and arena quests) once that feature is implemented. In the meantime, since they are hardcoded, the event quest names in the Greeter menu are hardcoded (with the exception of the Jhen event quest, which only appears if there is a sandstorm)

### DIFF
--- a/mh/constants.py
+++ b/mh/constants.py
@@ -21,10 +21,11 @@
 
 import struct
 from other.utils import pad
-from mh.time_utils import current_tick, TICKS_PER_CYCLE, get_jhen_event_times
+from mh.time_utils import current_tick, TICKS_PER_CYCLE, get_jhen_event_times,\
+    is_jhen_active
 from mh.quest_utils import QUEST_EVENT_JUMP_FOUR_JAGGI, QUEST_EVENT_BLOOD_SPORT,\
     QUEST_EVENT_MERCY_MISSION, QUEST_EVENT_THE_PHANTOM_URAGAAN, QUEST_EVENT_WORLD_EATER,\
-    QUEST_EVENT_WHERE_GODS_FEAR_TO_TREAD, QUEST_EVENT_GREEN_EGGS
+    QUEST_EVENT_WHERE_GODS_FEAR_TO_TREAD, QUEST_EVENT_GREEN_EGGS, get_quest_name
 
 
 def make_binary_type_time_events():
@@ -164,13 +165,45 @@ def make_binary_npc_greeters(is_jap=False):
     US_OFFSET = 0x180
     JP_OFFSET = 0x100
     offset = JP_OFFSET if is_jap else US_OFFSET
+
+
+    if is_jhen_active():
+        tool_shop = b"Half-off sale!"
+        material_shop = b"Half-off sale!"
+        event_quests = b"The Festival of Fear\n"
+    else:
+        tool_shop = b"Prices are normal."
+        material_shop = b"Prices are normal."
+        event_quests = b""
+
+    event_quests += b"\n".join([
+        get_quest_name(QUEST_EVENT_JUMP_FOUR_JAGGI),
+        get_quest_name(QUEST_EVENT_BLOOD_SPORT),
+        get_quest_name(QUEST_EVENT_MERCY_MISSION),
+        get_quest_name(QUEST_EVENT_THE_PHANTOM_URAGAAN),
+        get_quest_name(QUEST_EVENT_WORLD_EATER),
+        get_quest_name(QUEST_EVENT_WHERE_GODS_FEAR_TO_TREAD),
+        get_quest_name(QUEST_EVENT_GREEN_EGGS)
+    ])
+
+    guildmaster_str = (
+        b'To all hunters:\n'
+        b'Loc Lac is in "beta,"\n'
+        b'whatever that means. Also,\n'
+        b'the Arena is currently under\n'
+        b'construction. And now, a:\n'
+        b'haiku: "Patient warriors /\n'    
+        b'returning home at long last /\n'
+        b'now join hands to hunt!"'
+    )
+
     data = b""
-    data += pad(b"Plaza Tool Shop\n\nNot supported yet.", offset)
-    data += pad(b"Material shop unavailable\nyet.", offset)
-    data += pad(b"Trading post closed at\nthe moment.", offset)
-    data += pad(b"No event quests.", offset)
+    data += pad(tool_shop, offset)
+    data += pad(material_shop, offset)
+    data += pad(b"The Trading Post is open!", offset)
+    data += pad(event_quests, offset)
     data += pad(b"No arena quests.", offset)
-    data += pad(b"To all hunters:\n\nThis is a test server.", offset)
+    data += pad(guildmaster_str, offset)
     return data
 
 
@@ -260,7 +293,7 @@ PAT_BINARIES = {
     },
     0x03: {
         "version": 1,
-        "content": make_binary_npc_greeters(is_jap=IS_JAP)
+        "content": lambda: make_binary_npc_greeters(is_jap=IS_JAP)
     },
     0x04: {
         "version": 1,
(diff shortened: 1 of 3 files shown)


InusualZ (approved):
LGTM

mh/constants.py
```diff
@@ -164,13 +165,41 @@ def make_binary_npc_greeters(is_jap=False):
...
+
+
+    if is_jhen_active():
+        tool_shop_str = "Half-off sale!"
+        material_shop_str = "Half-off sale!"
+        event_quests_str = "The Festival of Fear\n"
+    else:
+        tool_shop_str = "Prices are normal."
+        material_shop_str = "Prices are normal."
+        event_quests_str = ""
```
sepalani, line 177: Use bytes literals (`b""` so you don't have to encode the strings).

mh/constants.py
```diff
@@ -164,13 +165,41 @@ def make_binary_npc_greeters(is_jap=False):
...
+
+    event_quests_str += QUEST_EVENT_JUMP_FOUR_JAGGI[0:40].decode('ascii').rstrip('\x00') + '\n'
+    event_quests_str += QUEST_EVENT_BLOOD_SPORT[0:40].decode('ascii').rstrip('\x00') + '\n'
+    event_quests_str += QUEST_EVENT_MERCY_MISSION[0:40].decode('ascii').rstrip('\x00') + '\n'
+    event_quests_str += QUEST_EVENT_THE_PHANTOM_URAGAAN[0:40].decode('ascii').rstrip('\x00') + '\n'
+    event_quests_str += QUEST_EVENT_WORLD_EATER[0:40].decode('ascii').rstrip('\x00') + '\n'
+    event_quests_str += QUEST_EVENT_WHERE_GODS_FEAR_TO_TREAD[0:40].decode('ascii').rstrip('\x00') + '\n'
+    event_quests_str += QUEST_EVENT_GREEN_EGGS[0:40].decode('ascii').rstrip('\x00')
```
sepalani, line 185: You should create a helper function which extracts the quest's name:
```python
event_quests_str = b"\n".join([
    get_quest_name(QUEST_1),
    get_quest_name(QUEST_2),
    ...
])
```

mh/constants.py
```diff
@@ -164,13 +165,41 @@ def make_binary_npc_greeters(is_jap=False):
...
+
+    guildmaster_str = "To all hunters:\n"
+    guildmaster_str += "Loc Lac is in \"beta,\"\n"
+    guildmaster_str += "whatever that means. Also,\n"
+    guildmaster_str += "the Arena is currently under\n"
+    guildmaster_str += "construction. And now, a:\n"
+    guildmaster_str += "haiku: \"Patient warriors /\n"    
+    guildmaster_str += "returning home at long last /\n"
+    guildmaster_str += "now join hands to hunt!\""
```
sepalani, line 194: The following should work and get rid of the concatenation:
```python
guildmaster_str = (
    b'To all hunters:\n'
    b'Loc Lac is in "beta",\n'
    b'whatever that means. Also,\n'
    b'the Arena is currently under\n'
    b'construction. And now, a:\n'
    b'haiku: "Patient warriors /\n'
    b'returning home at long last /\n'
    b'now join hands to hunt!"'
)
```

mh/time_utils.py
```diff
@@ -87,6 +87,11 @@ def get_jhen_event_times():
             int(cycle_start + JHEN_END*SECONDS_PER_DAY))  # sandstorm end
 
 
+def is_jhen_active():
+    day_in_cycle = int(current_server_time()//SECONDS_PER_DAY) % JHEN_EVENT_OFFSET
+    return day_in_cycle >= JHEN_START and day_in_cycle < JHEN_END
```
sepalani, line 92:
```python
return JHEN_START <= day_in_cycle < JHEN_END
```

sepalani (changes requested):

mh/constants.py
kbarkevich: Done, except for event quests due to the `get_quest_name` returning a normal string

mh/constants.py
kbarkevich: Done

mh/constants.py
kbarkevich: Done

mh/time_utils.py
kbarkevich: Done

mh/constants.py
sepalani: `b"\n".join` is missing and `get_quest_name` should return `bytes`.

mh/constants.py
sepalani: If my comment on `get_quest_name` is addressed, that's doable.

mh/quest_utils.py
```diff
@@ -1325,6 +1325,9 @@ def read_quest_sm_data(fname):
         data = f.read()
     return bytearray.fromhex(data.replace(" ","").replace("\n",""))
 
+def get_quest_name(quest):
+    return quest[0:40].decode('ascii').rstrip('\x00')
```
sepalani, line 1329: This should return `bytes` so we don't have to handle the encoding. There are other cleaner ways to get a null-terminated string from a buffer of bytes (for instance, using the `ctypes` module).

If the string is guaranteed to be null terminated:
```python
return ctypes.create_string_buffer(quest).value
```
Otherwise, if it has a fixed size and isn't guaranteed to be null terminated:
```python
return ctypes.create_string_buffer(quest[:40]).value
```

There is another way involving slices and `bytes.find` but it will be cumbersome if you have to deal with cases where the string isn't null-terminated.

sepalani (changes requested):

mh/constants.py
```diff
@@ -164,13 +165,43 @@ def make_binary_npc_greeters(is_jap=False):
...
+
+    event_quests_str += get_quest_name(QUEST_EVENT_JUMP_FOUR_JAGGI) + '\n'
+    event_quests_str += get_quest_name(QUEST_EVENT_BLOOD_SPORT) + '\n'
+    event_quests_str += get_quest_name(QUEST_EVENT_MERCY_MISSION) + '\n'
+    event_quests_str += get_quest_name(QUEST_EVENT_THE_PHANTOM_URAGAAN) + '\n'
+    event_quests_str += get_quest_name(QUEST_EVENT_WORLD_EATER) + '\n'
+    event_quests_str += get_quest_name(QUEST_EVENT_WHERE_GODS_FEAR_TO_TREAD) + '\n'
+    event_quests_str += get_quest_name(QUEST_EVENT_GREEN_EGGS)
+
+    guildmaster_str = (
```
sepalani, line 187: The `_str` suffix in the variable names should be removed. It's superfluous and doesn't reflect that these should be `bytes`. Moreover, the function is short and mostly self-contained with an explicit name `npc_greeters` so I don't think more context is needed to understand the variables purpose.

sepalani (changes requested):

mh/constants.py
kbarkevich: Changed

mh/quest_utils.py
kbarkevich: Changed

mh/constants.py
kbarkevich: Fixed

mh/constants.py
kbarkevich: Done

mh/constants.py
```diff
@@ -21,10 +21,11 @@
 
 import struct
 from other.utils import pad
-from mh.time_utils import current_tick, TICKS_PER_CYCLE, get_jhen_event_times
+from mh.time_utils import current_tick, TICKS_PER_CYCLE, get_jhen_event_times,\
+    current_server_time, is_jhen_active
```
sepalani, line 25: Is importing `current_server_time` still necessary?

sepalani (changes requested):
Otherwise, LGTM.

mh/constants.py
kbarkevich: Fixed

sepalani (approved):
Tested and working on both the western and japanese versions.

PR #79, opened 2022-12-08
Fixed bug with incorrect player count in Servers

kbarkevich:
Re-implemented logic based on the byte sent with the `ReqShut` packet, which, if 1, instructs the server to disconnect the player completely. It differentiates between when the client is switching from one server to another, and when the client is fully disconnecting from the game.

This fixes the bug where, when players disconnected from the screen that displays the list of gates in a server, they would not be removed from the list of players in the server. This resulted in an incorrect player count in the affected server.


sepalani (approved):
LGTM, untested.

PR #80, opened 2022-12-11
Fixed time divergence issue

kbarkevich:
This PR fixes the issue where players who log in later than others have their times initialized ahead of everybody else.

This was caused due to the "countdown" that was previously used to advance time. With the addition of the time-based event quests, the correct timer now progresses forwards, making the countdown unneeded. The countdown has been replaced with a single variable set upon server initialization to the timestamp at which the server was booted up. As a consequence, the in-game day/night cycle will always start at the beginning of daytime when the server boots up.

If we wished to make the day/night cycle continuous even across periods when the server was off, like it was before, we could set the value to a constant timestamp instead.

### DIFF
--- a/mh/constants.py
+++ b/mh/constants.py
@@ -21,7 +21,7 @@
 
 import struct
 from other.utils import pad
-from mh.time_utils import current_tick, TICKS_PER_CYCLE, get_jhen_event_times,\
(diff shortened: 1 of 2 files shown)


sepalani (approved):
LGTM. I guess not setting the event time provoked an underflow somewhere.

PR #82, opened 2022-12-15
Players class: added Lock and improved thread-safety of related code

kbarkevich:
- Added parent `Lockable` class for future thread-safe classes
- `Players` class now inherits from `Lockable` class
- Added locking guards to `Players` class methods that alter the state of the object
- Added `Players` locking guards to `pat` handlers that request the state of a `Players` object over multiple steps

### DIFF
--- a/fmp_server.py
+++ b/fmp_server.py
@@ -139,16 +139,17 @@ class FmpRequestHandler(PatRequestHandler):
         TR: Layer sync user list response
         """
         players = self.session.get_layer_players()
-        count = len(players)
-        data = struct.pack(">I", count)
+        with players.lock():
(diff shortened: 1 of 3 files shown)


sepalani (changes requested):
IMO, `"PLAYER LOCKING"` comments shouldn't be added to the docstring and we should expose a member method or variable such as `.lock()` to make the intent of the `with` statement obvious. We might also want to provide a `DummyLockable` implementation since a lock isn't needed in some circumstances where there is no concurrency (i.e. single-thread) and when it's handled by the database backend instead.

Otherwise, the code seems sane to me.

PR #85, opened 2023-02-02
Circle class: added Lock and improved thread-safety of related code

kbarkevich:
`Circle` class now inherits from the `Lockable` class.

Added locking (using `with` blocks) sections to improve thread safety.

### DIFF
--- a/fmp_server.py
+++ b/fmp_server.py
@@ -365,9 +365,10 @@ class FmpRequestHandler(PatRequestHandler):
 
         data = b''
         for i, circle in enumerate(city.circles):
-            if not circle.is_empty():
-                data += pati.CircleInfo.pack_from(circle, i+1)
-                count += 1
+            with circle.lock():
+                if not circle.is_empty():
+                    data += pati.CircleInfo.pack_from(circle, i+1)
+                    count += 1
 
         data = struct.pack(">II", unk, count) + data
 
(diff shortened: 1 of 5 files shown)


mh/pat_item.py
```diff
@@ -693,61 +693,62 @@ class CircleInfo(PatData):
 
     @staticmethod
     def pack_from(circle, circle_index):
-        circle_info = CircleInfo()
-        circle_info.index = Long(circle_index)
+        with circle.lock():
```
sepalani, line 696: Can't it be locked where it's going to be packed rather than in the `pack_from` function?

mh/session.py
```diff
@@ -309,16 +309,18 @@ def try_transfer_circle_leadership(self):
             return None, None
 
         circle = self.get_circle()
-        if circle.leader != self or circle.get_population() <= 1 or not circle.departed:
+        with circle.lock():
```
sepalani, line 312: Can't you lock both at the same time?
```python
with circle.lock() and circle.player.lock():
```

fmp_server.py
```diff
@@ -388,37 +389,37 @@ def recvReqCircleCreate(self, packet_id, data, seq):
 
         city = self.session.get_city()
         circle, circle_index = city.get_first_empty_circle()
+        with circle.lock():
```
sepalani, line 392: The `with` statement should be moved closer to its use, just above:
> circle.leader = self.session

fmp_server.py
```diff
@@ -585,15 +586,17 @@ def sendAnsCircleUserList(self, seq):
...
-            circle_user_data = pati.CircleUserData()
-            circle_user_data.is_standby = pati.Byte(
-                int(player.is_circle_standby()))
-            circle_user_data.player_index = pati.Byte(i+1)
-            circle_user_data.capcom_id = pati.String(player.capcom_id)
-            circle_user_data.hunter_name = pati.String(player.hunter_name)
-            data += circle_user_data.pack()
+        with circle.lock():
+            data = struct.pack(">I", circle.get_population())
+            with circle.players.lock():
```
sepalani, line 591:
```python
with circle.lock() and circle.players.lock():
```

sepalani (changes requested):
Here is a cursory review.

mh/pat_item.py
kbarkevich: All situations where the `pack_from` method is called will require a lock over the whole thing, so wouldn't it be simpler to just have the `pack_from` methods themselves contain the lock? That way we don't need to worry about doing the locking every time we call these methods.

mh/session.py
kbarkevich: Done, only using a comma rather than an `and`, as per https://docs.python.org/3/reference/compound_stmts.html#the-with-statement

fmp_server.py
kbarkevich: Done

fmp_server.py
kbarkevich: Done, using comma as described above

mh/pat_item.py
InusualZ: I agree with @kbarkevich, maybe making the locking more explicit? Another method called: `lock_and_pack_from`

sepalani: The current approach should be safe as the lock is reentrant. I'll merge it as is for the time being.

sepalani (approved):

PR #87, opened 2023-02-03
PAT: Implement NtcLayerMediation packets

sepalani:
This PR implements the NtcLayerMediation packets properly. At the moment, I kept them disabled since the city doesn't retain the mediation list and it can produce oddities upon player disconnection or player entering the city.

Ready to be reviewed & merged.

### DIFF
--- a/mh/pat.py
+++ b/mh/pat.py
@@ -2401,16 +2401,17 @@ class PatRequestHandler(server.BasicPatHandler):
         ID: 64820200
         JP: レイヤ調停データリスト取得返答
         TR: Get layer mediation list response
+
+        TODO: Store the item list and its state in the city.
         """
         unk = 0
-        count = 1
-        item = pati.MediationListItem()
-        item.name = pati.String(b"Test")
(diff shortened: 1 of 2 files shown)


InusualZ (commented):
I'm sorry 🤐 I can't remember what does `LayerMediation` mean or when is this packet used?

Also, it may be worth to implement it with the memory-db that we currently have. Even if just to have a proof of concept.

sepalani:
The `LayerMediation` packets are related to the ownership of seat or the barrel (arm wrestling). Using these Ntc packets allow to disable the cursor of an already owned seat (and prevent multiple hunters to take the same seat).

My issue with enabling it is that will need to handle player disconnection and other side effects that might "lock" a seat if not done properly. It'd require proper testing before I'd be confident to integrate it into the memory-db. I'd also like to add a proper check tight to the city, so we don't always attribute the resource when hunters ask for one.

Having these packets working is just a first step, before we can achieve that I believe.

InusualZ (approved):
Make sense and code lgtm

PR #90, opened 2023-02-21
Added locking mechanism to City class

kbarkevich:
Extension of the `Locking` class inheritance logic to the `City` class.

### DIFF
--- a/mh/database.py
+++ b/mh/database.py
@@ -221,7 +221,7 @@ class Circle(Lockable):
             self.unk_byte_0x0e = 0
 
 
-class City(object):
+class City(Lockable):
     LAYER_DEPTH = 3
 
     def __init__(self, name, parent):
@@ -236,6 +236,7 @@ class City(object):
             # One circle per player
             Circle(self) for _ in range(self.get_capacity())
         ]
+        super(City, self).__init__()
 
     def get_population(self):
         return len(self.players)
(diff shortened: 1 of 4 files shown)


kbarkevich:
Rebased onto dev

mh/session.py
```diff
@@ -190,7 +190,8 @@ def layer_down(self, layer_id):
     def layer_create(self, layer_id, settings, optional_fields):
         if self.layer == 1:
             city = self.create_city(layer_id, settings, optional_fields)
-            city.leader = self
+            with city.lock():
```
InusualZ, line 193: I don't think that you need to lock the city that you just created. 

Now that I think about it, why not move the assignment of the leader to the `create_city` method?

mh/pat.py
```diff
@@ -84,14 +84,17 @@ def layer_broadcast(self, session, packet_id, data, seq,
...
+        with circle.lock(), circle.players.lock():
             for _, player in circle.players:
                 if session and player == session:
                     continue
 
                 handler = self.get_pat_handler(player)
                 if handler:
-                    handler.try_send_packet(packet_id, data, seq)
+                    handlers.append(handler)
+        for handler in handlers:
```
InusualZ, line 96: I suppose you do this dance of collecting the handlers first and then send the payload too avoid a deadlock?

mh/database.py
```diff
@@ -579,36 +585,40 @@ def find_users(self, capcom_id="", hunter_name=""):
...
+        with city.lock():
+            city.optional_fields = optional_fields
         return city
 
     def join_city(self, session, server_id, gate_id, index):
         city = self.get_city(server_id, gate_id, index)
-        city.parent.players.remove(session)
-        city.players.add(session)
+        with city.lock():
+            city.parent.players.remove(session)
```
InusualZ, line 595: Why aren't we locking the `parent.players`? Just curious, just want to know your reasoning, in any case I don't think it's a big deal.

InusualZ (commented):
Overall locking seems to be at the right places. Mostly questions.

mh/session.py
kbarkevich: All writes should have locks on them. However I think you're right, it would be better to have the city leader be assigned inside the `create_city` method. After all, no cities can exist without leaders.

mh/pat.py
kbarkevich: That's part of it. It's mostly that it's bad practice to dispatch packets (or do any unnecessarily lengthy operation) with a lock in place- it slows down the program. Better to handle the packet dispatching once the lock is terminated.

I have a separate branch (based off this one, so no PR yet) in which I generalized this concept into a PacketQueue class that is entered using a `with` statement alongside locks that have packet dispatchers inside them. That will enforce this "no packet sending until the lock is done" rule throughout the server.

mh/database.py
kbarkevich: `Players.remove()` and `Players.add()` already have locks inside of them, but you're right. Technically they should be unified here to ensure the lock is not dropped between player removal and player addition. That is out of scope of this PR though, being solely related to the `Players` class, so if it is to be fixed it should happen in another separate PR.

mh/session.py
kbarkevich: Changed this and pushed.

sepalani (approved):
Sounds reasonable.

PR #91, opened 2023-02-24
Typo fix

jo9182:
This has been bugging me.


sepalani:
Good catch. Though, I don't plan to merge it to master since we'll rewrite the whole README when will release the beta update. Thank you for your understanding.

PR #93, opened 2023-03-11
Implemented Warping

kbarkevich:
This PR handles warping in the following cases:

* Same Gate, City->City
* Same Gate, Gate->City
* Same Gate, City->Gate
* Same Server, Different Gate, City->City
* Same Server, Different Gate, Gate->City
* Same Server, Different Gate, City->Gate
* Same Server, Different Gate, Gate->Gate
* Different Server, City->City
* Different Server, Gate->City
* Different Server, City->Gate
* Different Server, Gate->Gate

"Different Server" warps are a multi-step process that periodically checks to see if the destination is full (or empty, in the case of a city.) The steps are as follows:

1. `RecLayerJump`/`AnsLayerJump` (Game initiates warp. First check to see if destination is valid happens.)
2. `ReqLayerJumpReady`/`NtcLayerJumpReady` (Game asks to initiate Server->Server hop, does so using standard FMP disconnect/connect sequences upon reply with `Ntc` packet. Second check to see if destination is valid happens.)
3. `ReqLayerJumpGo`/`NtcLayerJumpGo` (Game, now having completed the Server->Server hop, asks to enter their destination Gate/City. Third check to see if destination is valid happens.)

Testing has shown that the process is able to be gracefully aborted through an `AnsAlert` packet at any step. For example, if the Server->Server jump is completed, and then the process is aborted, the user will be returned to the "server select" screen. If the process is aborted before the Server->Server jump is completed, the user will remain in their current location.

TODO in a future PR: Implementation of `Gate` locking at the relevant steps. Currently, `City` locking is implemented.

### DIFF
--- a/fmp_server.py
+++ b/fmp_server.py
@@ -40,6 +40,237 @@ class FmpRequestHandler(PatRequestHandler):
         self.server.debug("Connection: {!r}".format(connection_data))
         self.sendNtcLogin(3, connection_data, seq)
 
+    def recvReqLayerJump(self, packet_id, data, seq):
+        """ReqLayerJump packet.
+        
+        ID: 64100100
+        JP: レイヤ移動要求（位置バイナリ）
+        TR: Layer move request (position binary)
+        
+        A relay of the target user's layer_host.
+        """
+        length, unk1, server_id, unk2, gate_id, city_id = struct.unpack_from(">HIIHHH", data)
+
+        if self.session.local_info["server_id"] == server_id and\
+            self.session.local_info["gate_id"] == gate_id and\
+            self.session.local_info["city_id"] == city_id:
+            self.sendAnsAlert(PatID4.AnsLayerJump,
+                        "<LF=8><BODY><CENTER>You're already there.<END>",
+                        seq)
+        elif self.session.local_info["server_id"] != server_id:
+            # Different Server
+            if self.session.get_server_full(server_id) or\
+                self.session.get_gate_full(server_id, gate_id) or\
+                (city_id and self.session.get_city_full(server_id, gate_id, city_id)):
+                self.sendAnsAlert(PatID4.AnsLayerJump,
+                        "<LF=8><BODY><CENTER>Destination is full.<END>",
+                        seq)
+                return
+        else:
+            # TODO: Lock the gate before the player checks its capacity.
+            # with self.session.find_gate(server_id, gate_id).lock()
+            if self.session.local_info["gate_id"] != gate_id:
+                # Same Server, different Gate
+                if city_id:
+                    city = self.session.find_city(server_id, gate_id, city_id)
+                    with city.lock():
+                        if city.is_empty():
+                            self.sendAnsAlert(PatID4.AnsLayerJump,
+                                    "<LF=8><BODY><CENTER>City no longer exists.<END>",
+                                    seq)
+                            return
+                        elif self.session.get_gate_full(server_id, gate_id) or\
+                            city.is_full():
+                            self.sendAnsAlert(PatID4.AnsLayerJump,
+                                    "<LF=8><BODY><CENTER>Destination is full.<END>",
+                                    seq)
+                            return
+
+                        # Step 1: Leave current location
+                        while self.session.layer > 0:
+                            self.notify_layer_departure()
+                            self.session.layer_up()
+                        # Step 2: Join Gate
+                        self.session.layer_down(gate_id)
+                        user = pati.LayerUserInfo()
+                        user.capcom_id = pati.String(self.session.capcom_id)
+                        user.hunter_name = pati.String(self.session.hunter_name)
+                        user.stats = pati.Binary(self.session.hunter_info.pack())
+
+                        data = pati.lp2_string(self.session.capcom_id)
+                        data += user.pack()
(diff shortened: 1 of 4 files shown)


mh/session.py
```diff
@@ -139,6 +139,24 @@ def get_server(self):
...
+        return DB.get_city(server_id, gate_id, city_id).is_full()
+
+    def get_city_empty(self, server_id, gate_id, city_id):
+        return DB.get_city(server_id, gate_id, city_id).is_empty()
+
+    def find_gate(self, server_id, gate_id):
+        return DB.get_gate(server_id, gate_id)
+
+    def find_city(self, server_id, gate_id, city_id):
+        return DB.get_city(server_id, gate_id, city_id)
```
sepalani, line 158: I'm really not a huge fan of this since the `session` module is mainly used to cache the server data and hide the implementation logic _(for instance which DB backend we use, which function to call based on the packet we received and its parameters, etc.)_. That's also why most of its functions don't ask for the `server_id` as parameter and use the `local_info` when needed rather than being a simple wrapper to call the default database instance.

I'll let it slide for now since we don't provide a better abstraction to query the database directly, especially for other servers. AFAICT, this point seems still valid with the `state` abstraction you proposed in another PR because parts of the `session` modules (including the ones this PR added) needs to take into account whether or not the city/gate/server is handled by the current server process or not.

fmp_server.py
```diff
@@ -40,6 +40,237 @@ def recvAnsConnection(self, packet_id, data, seq):
...
+                        self.notify_layer_departure()
+                        self.session.layer_up()
+        self.sendAnsLayerJump(seq)
+
+    def sendAnsLayerJump(self, seq):
+        """NtcLayerJumpGo packet.
+
+        ID: 64170200
+        JP: レイヤ予約移動実行返答
+        TR: Layer reservation move execution reply
```
sepalani, line 162:
```suggestion
    def sendAnsLayerJump(self, seq):
        """AnsLayerJump packet.

        ID: 64100200
        JP: レイヤ移動返答（位置バイナリ）
        TR: Layer move reply (position binary)
```

fmp_server.py
```diff
@@ -40,6 +40,237 @@ def recvAnsConnection(self, packet_id, data, seq):
...
+        """
+        self.send_packet(PatID4.AnsLayerJump, b"", seq)
+
+    def recvReqLayerJumpReady(self, packet_id, data, seq):
+        """AnsLayerJump packet.
+
+        ID: 64100200
+        JP: レイヤ移動返答（位置バイナリ）
+        TR: Layer move reply (position binary)
```
sepalani, line 171:
```suggestion
    def recvReqLayerJumpReady(self, packet_id, data, seq):
        """ReqLayerJumpReady packet.

        ID: 64160100
        JP: レイヤ予約移動確認要求
        TR: Layer reservation move confirmation request
```

fmp_server.py
```diff
@@ -40,6 +40,237 @@ def recvAnsConnection(self, packet_id, data, seq):
...
+            self.session.layer_up()
+        self.sendNtcLayerJumpReady(seq)
+
+    def sendNtcLayerJumpReady(self, seq):
+        """ReqLayerJumpReady packet.
+
+        ID: 64160200
+        JP: レイヤ予約移動確認返答
+        TR: Layer reservation move confirmation reply
+        """
```
sepalani, line 196:
```suggestion
    def sendNtcLayerJumpReady(self, seq):
        """NtcLayerJumpReady packet.

        ID: 64160200
        JP: レイヤ予約移動確認返答
        TR: Layer reservation move confirmation reply
        """
```
sic: I think this function was wrongly identified when I generated the constants so we should rename it `AnsLayerJumpReady` later at some point or add a comment stating that this function was wrongly named in game if that was the case.

fmp_server.py
```diff
@@ -40,6 +40,237 @@ def recvAnsConnection(self, packet_id, data, seq):
...
+                                        data, seq)
+        self.sendNtcLayerJumpGo(seq)
+
+    def sendNtcLayerJumpGo(self, seq):
+        """NtcLayerJumpGo packet.
+
+        ID: 64170200
+        JP: レイヤ予約移動実行返答
+        TR: Layer reservation move execution reply
+        """
```
sepalani, line 271: sic: Ditto, regarding its name. _Feel free to ignore this comment._

sepalani (changes requested):
This is my first pass on this PR. I'll try to have a look at these `if`/`elif`/`else` to find a cleaner way to refactor it.

fmp_server.py
```diff
@@ -40,6 +40,237 @@ def recvAnsConnection(self, packet_id, data, seq):
...
         self.sendNtcLogin(3, connection_data, seq)
 
+    def recvReqLayerJump(self, packet_id, data, seq):
+        """ReqLayerJump packet.
+        
+        ID: 64100100
+        JP: レイヤ移動要求（位置バイナリ）
+        TR: Layer move request (position binary)
+        
+        A relay of the target user's layer_host.
```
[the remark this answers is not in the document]
InusualZ: I was thinking about that, and my idea was that they would have a method of communicating to the other server to reserve a slot. That way you avoid having an issue of state changing mid flight. But I don't if they actually did that or if it would be fast enough to not create a noticeable slowdown in the process.

PR #95, opened 2023-03-18
Calculate the amount of gate and city based on the server's capacity

InusualZ:
Exactly as the title say, compute the amount of city and gates based on the server's capacity.

### DIFF
--- a/mh/database.py
+++ b/mh/database.py
@@ -19,6 +19,7 @@
     along with this program.  If not, see <http://www.gnu.org/licenses/>.
 """
 
+import math
 import random
 import time
 from threading import RLock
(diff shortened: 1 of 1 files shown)


mh/database.py
```diff
@@ -293,16 +298,23 @@ def reserve(self, reserve):
...
-    def __init__(self, name, parent, city_count=40, player_capacity=100):
+    def __init__(self, name, parent, capacity=100, is_jap=False):
         self.name = name
         self.parent = parent
         self.state = LayerState.EMPTY
+
+        city_max_capacity = City.get_max_capacity(is_jap)
+        city_count = int(min(
+            math.ceil(capacity / city_max_capacity), Gate.MAX_CITY_PER_GATE))
+        capacity = city_count * city_max_capacity
```
sepalani, line 311: Won't this approach fail when many users decide to solo play (and not fill up the city)?

InusualZ: yes, but even in that case there is not much that we can do, because there is a maximum amount of city per gate.

PR #98, opened 2023-05-07
Database: Add TempSQLiteDatabase class

sepalani:
This PR adds an hybrid database based on the temporary database and a sqlite database. It provides a way to preserve the saved data between shutdowns. Currently, it preserves the data used by the following DB interfaces: `use_user`, `accept_friend` and `delete_friend`.

This PR also provides a thread-safe wrapper to workaround the SQLite concurrency limitations. I also used `setdefault` in the `DebugDatabase` to only set the default value if it doesn't exist.

Some remarks regarding my current approach:
 - _(the most important one)_ Only the connection should be behind this thread_local wrapper. Otherwise, the database class and its data will be duplicated which means that each thread will have its own independent database with its (inconsistent) data.
 - The current approach using a singleton-like `get_instance` isn't thread safe since we're setting it in a module's variable (similar to a `static global` variable in C) and this variable is accessed by the `session` module once and might be used by many threads. In other words, I can't (easily/cleanly) instantiate a new database object when a new thread is spawned.
 - That's why I chose this wrapper approach and used a `thread_local` context since it doesn't depend on how/when/where the objects are instantiated.

Tested on both Python 2/3 and MH3 USA/EUR/JAP.

### DIFF
--- a/mh/database.py
+++ b/mh/database.py
@@ -20,8 +20,10 @@
 """
 
 import random
+import sqlite3
 import time
-from threading import RLock
+from threading import RLock, local as thread_local
+
 
 CHARSET = "0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZ"
 
@@ -31,6 +33,8 @@ MEDIA_VERSIONS = {
     "1001301045": "RMHE08",
     "1002121503": "RMHP08",
 }
+
+BLANK_CAPCOM_ID = "******"
 
 RESERVE_DC_TIMEOUT = 40.0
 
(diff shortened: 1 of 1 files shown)


mh/database.py
```diff
@@ -695,11 +696,194 @@ def get_friends(self, capcom_id, first_index=None, count=None):
...
+    The following data need to be retained after a shutdown:
+     - Online support code and its Capcom IDs
+     - Friend list per Capcom ID
+     - Properties (at least the hunter name) per Capcom ID
+
+    TODO:
+     - Need proper documentation, especially for the overridable interface
+    """
+
+    DATABASE_NAME = "tmp_database.db"
```
InusualZ, line 776: Maybe `mh3sp.db`?

InusualZ (approved):
Fairly simple implementation and easy to follow. I like it :)

sepalani:
@InusualZ 
Done.

PR #100, opened 2023-05-14
PEP8: Fix most issues from our files

sepalani:
This PR fixes PEP8 coding style issues. The purpose of this PR is to allow, in the future, the use of github actions and/or other CI/CD tools to check the code/PRs for issues.

@kbarkevich @InusualZ 
If the online event/arena quests don't match the ones of the closed-beta, do you mind to diff the quests from there against the dev branch so I can update this PR accordingly.

### DIFF
--- a/fmp_server.py
+++ b/fmp_server.py
@@ -367,7 +367,9 @@ class FmpRequestHandler(PatRequestHandler):
                 circle.remarks = pati.unpack_string(circle_info.remarks)
 
             if "unk_byte_0x0e" in circle_info:
-                circle.unk_byte_0x0e = pati.unpack_byte(circle_info.unk_byte_0x0e)
+                circle.unk_byte_0x0e = pati.unpack_byte(
+                    circle_info.unk_byte_0x0e
+                )
 
             circle.reset_players(pati.unpack_long(circle_info.capacity))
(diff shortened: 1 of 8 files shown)


InusualZ:
What do you think about increasing the line max width?

sepalani:
> What do you think about increasing the line max width?

When there is a real need for it I wouldn't mind. Though, at the time I don't think/feel the need to do so. For instance, we rarely have too many indent levels _(example: class + method + with clause + for/while/if statement -> 16 columns used / 63 columns left)_. 

So having about ~60 characters left sounds fine/reasonable for python code, IMHO. My stance would have been different for C# or Java code of course since the syntax is more verbose _(type declaration, package members/qualified names, etc.)_. When we do go beyond that 79 char limit, I often question myself if there is a cleaner way to write that code (and often there is). Moreover, it allows to have a human and machine-readable code (examples: git diff, code editors, etc.). Do you have examples where you'd have needed more width?

By looking at this PR changes, there are cases where lines were insanely long and even with modern IDE the soft-wrap _(which I'm not found of)_ doesn't help to grasp the code meaning. Ditto, for cases where many parameters are passed to a function.

InusualZ (approved):

PR #103, opened 2023-05-30
Add a client socket timeout

sepalani:
This should mitigate the random issues/blocking reads the server was having.


mh/server.py
```diff
@@ -273,6 +273,10 @@ def accept_new_connection(self):
             return
 
         try:
+            # TODO: Find a cleaner way to process ill-formed packets.
+            # Currently, they get stuck on `packet = selected.on_recv()`,
+            # thus blocking the `serve_forever` method.
+            client_socket.settimeout(5.0)
```
InusualZ, line 279: `5` seconds seems a bit too much? What about something like `2.5` or `3`. And even those seems bit too much to me. My reasoning is that, let's pretend that is **not** a ill-formed packet, but a laggy connection, should we really accommodate a connection that its packet are lagging 5 seconds?

InusualZ (approved):
Just a mild concern.

mh/server.py
sepalani: Fair point, I will set it to `2.0` then. Since this option only affects blocking operations and these operations are processed after `poll`/`select`.

PR #111, opened 2023-07-03
Added event quest rotation

kbarkevich:
This PR adds the foundation for an event/arena quest cycle, with a new list of event/arena quests able to be scheduled to be sent each day. Currently, the event cycle contains 14 unique temporal slots, for 14 days of available scheduling. This is in order to keep the schedule aligned with the Sandstorm event, which is also on a 14-day cycle.


InusualZ:
LGTM

I do think that we need to stop hard coding the quests in code. We should come up with a system that enable us to define quest from a external file (and maybe even hot reload?)

InusualZ (approved):

sepalani (changes requested):
I'm not a huge fan of the way this is handled as the "temporal slots" are hardcoded and hooking more things (and lambda) into a not flexible already hardcoded system doesn't feel clean at all.

In other words, we shouldn't hardcode quests, nor its rotation system, nor tight it to the jhen event. Ideally, we should be able to change these settings without rebooting the server.

I'm open to suggestions regarding how this can be implemented. I've been thinking about hosting binary quest files and using a config file to schedule which quest files should be served.

InusualZ:
I agree with you, but most if not all the requirement that you set would require a fairly robust config system. If we are going to "stop" the progression of game <-> server feature we should start asap trying to figure out the direction that we want to go

mh/constants.py
```diff
@@ -180,21 +185,16 @@ def make_binary_npc_greeters(is_jap=False):
...
-        GRUDGE_MATCH_TWO_FLAMES,
-        GRUDGE_MATCH_LAND_LORDS
-    )
+    for quest in binary_loader.get_quest_dicts(temporal_slot):
+        if quest is None:
+            continue
+        if is_arena_quest(generate_flags(*(quest['quest_info']['flags']))):
+            ARENA_LIST.append(quest)
+        else:
+            QUEST_LIST.append(quest)
```
sepalani, line 197: The quest loader should returns the `QUEST_LIST` and `ARENA_LIST` instead so the logic is kept there. Same remark regarding the binary loader so we don't need to pass it as parameter.

mh/constants.py
```diff
@@ -7,21 +7,26 @@
...
+def make_binary_type_time_events(binary_loader=None):
     return struct.pack(">III", *get_jhen_event_times())
 
 
+def make_event_slot(event):
+    return (make_binary_event_quest(event), event)
+
+
+def get_event_slot(slot_num, temporal_slot):
+    return lambda quest_loader: quest_loader[temporal_slot][slot_num] if slot_num < len(quest_loader[temporal_slot]) else b'\0' * 0x4B4
```
sepalani, line 23: Can't you create a function and return it instead? Moreover, this line is way too long.

mh/constants.py
```diff
@@ -326,8 +326,8 @@ def slot(item, qty):
         "content": make_binary_type_time_events
     },
     0x03: {
-        "version": 1,
-        "content": lambda: make_binary_npc_greeters(is_jap=IS_JAP)
+        "version": lambda binary_loader: current_event_time_slot() + 1,
+        "content": lambda temporal_slot: lambda binary_loader: make_binary_npc_greeters(binary_loader, is_jap=IS_JAP, temporal_slot=temporal_slot)
```
sepalani, line 330: Please provide functions/callbacks instead of lambdas to deduplicate the code and prevent the risk of mistakes. Ditto for the others below.

mh/constants.py
```diff
@@ -7,21 +7,26 @@
...
-    GRUDGE_MATCH_ROYAL_LUDROTH, GRUDGE_MATCH_BIRD_BRUTE, \
-    GRUDGE_MATCH_TWO_FLAMES, GRUDGE_MATCH_LAND_LORDS
-
-
-def make_binary_type_time_events():
+    get_jhen_event_times, is_jhen_active, current_event_time_slot
+from mh.quest_utils import make_binary_event_quest, generate_flags
+
+
+def make_binary_type_time_events(binary_loader=None):
```
sepalani, line 14: Is adding a useless optional parameter necessary?

mh/constants.py
```diff
@@ -148,7 +153,7 @@ def make_binary_server_type_list(is_jap=False):
     return data
 
 
-def make_binary_npc_greeters(is_jap=False):
+def make_binary_npc_greeters(binary_loader, is_jap=False, temporal_slot=current_event_time_slot()):
```
sepalani, line 156: `temporal_slot=current_event_time_slot()` will be evaluated as a default value once, is this behaviour intended?

mh/pat.py
```diff
@@ -1131,10 +1133,13 @@ def sendAnsBinaryHead(self, binary_type, seq):
...
-        content = binary["content"]
+        version = binary["version"]
+        if callable(version):
+            version = version(self.server.binary_loader)
+        content = get_pat_binary_from_version(binary_type, version)
         if callable(content):
-            content = content()
-        data = struct.pack(">II", binary["version"], len(content))
+            content = content(self.server.binary_loader)
+        data = struct.pack(">II", version, len(content))
```
sepalani, line 1142: Can't this logic/part be offloaded to the `binary_loader`?

mh/quest_utils.py
```diff
@@ -1548,921 +1745,63 @@ def read_quest_sm_data(fname):
...
+    elif isinstance(input, unicode):
+        return input.encode('utf-8')
+    else:
+        return input
+
+
+class QuestLoader:
+    def __init__(self, file_loc='event/quest_rotation.json'):
+        self.version = 0
+        self.version_incrementer = 28
```
sepalani, line 1767: Is this an arbitrary constant?

mh/quest_utils.py
```diff
@@ -1548,921 +1745,63 @@ def read_quest_sm_data(fname):
...
+        self.file_loc = file_loc
+        self.load_quests()
+        self.version = 0
+
+    def load_quests(self, file_loc=None):
```
sepalani, line 1772: Is `file_loc` ever used?

mh/quest_utils.py
```diff
@@ -1548,921 +1745,63 @@ def read_quest_sm_data(fname):
...
+                for quest_file in day:
+                    with open(quest_file, 'r') as ff:
+                        curr_quest = byteify(json.load(ff))
+                        curr_quest = make_event_slot(curr_quest)
+                        curr_quest_list.append(curr_quest)
+                quest_list.append(curr_quest_list)
+        self.quests = quest_list
+    
+    def get_quest(self, idx):
+        return self.quests(idx)
```
sepalani, line 1789: - Is this ever used?
- Isn't `self.quests` a list rather than a callable?

mh/quest_utils.py
```diff
@@ -1548,921 +1745,63 @@ def read_quest_sm_data(fname):
...
+    def __len__(self):
+        # Get number of days
+        return len(self.quests)
+
+    def __getitem__(self, idx):
+        # Get quests for the day
+        trueIdx = idx - self.version - 1
+        print("idx is {}".format(trueIdx))
+        print("idx: {}, len is {}".format(trueIdx, len(self.quests[trueIdx])))
+        return [self.quests[trueIdx][j][0] for j in range(len(self.quests[trueIdx]))]
```
sepalani, line 1805: - Please use `snake_case`.
- Can't `enumerate` be used instead of `range(len(...))`?
- These debug prints should be removed.

sepalani (changes requested):
Some comments regarding the code. Could you please also address the linter issues.

mh/quest_utils.py
kbarkevich: This gives us the ability to expand the quest rotation to up to 4 weeks (7 * 4 days) per update.

other/utils.py
```diff
@@ -340,7 +340,7 @@ def wii_ssl_wrap_socket(sock, ssl_cert, ssl_key):
     return context.wrap_socket(sock, server_side=True)
 
 
-def create_server(server_class, server_handler,
+def create_server(server_class, server_handler, binary_loader,
```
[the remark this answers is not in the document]
kbarkevich: It's currently initialized a single time in `master_server.py` and a reference to that object is passed down to the children that require it. Since it's stateful (currently in that it tracks an incrementing version number, and in the future in that we might have it on an auto refresh timer to check for updated files periodically) I'd rather do it this way.

dump_quest.py
```diff
@@ -0,0 +1,2937 @@
+"""Quest dumper.
```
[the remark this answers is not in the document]
kbarkevich: Fixed

dump_quest.py
```diff
@@ -0,0 +1,2937 @@
...
+    BowgunFrame, BowgunStock, BowgunBarrel
+
+def EXPORT(quest):
+    import json
+    name = quest['quest_info']['name'].replace(':', '')
+    with open('event/{}.json'.format(name), "w") as outfile:
+        json.dump(quest, outfile, indent=4)
+
+
+QUESTS = []
```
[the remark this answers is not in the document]
kbarkevich: This has been fixed

dump_quest.py
```diff
@@ -0,0 +1,2937 @@
...
+             (ItemsType.wyvernfire_lv1, 10), (ItemsType.water_s, 60),
+             (ItemsType.sleep_s_lv2, 8), (ItemsType.dragon_s, 20),
+             (ItemsType.demon_s_ii, 5))
+        )
+    )
+}
+QUESTS.append(GRUDGE_MATCH_LAND_LORDS)
+
+for q in QUESTS:
+    EXPORT(q)
```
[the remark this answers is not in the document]
kbarkevich: This has been fixed, as with above

master_server.py
```diff
@@ -32,8 +34,9 @@ def create_servers(silent=False, debug_mode=False):
 def main(args):
     """Master server main function."""
     register_debug_signal()
-
-    servers, has_ui = create_servers(silent=args.silent,
+    binary_loader = QuestLoader("event/quest_rotation.json")
```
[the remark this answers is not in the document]
kbarkevich: See reply to that comment, as it pertains to this as well

mh/quest_utils.py
```diff
@@ -1548,921 +1745,63 @@ def read_quest_sm_data(fname):
...
+def byteify(input):
+    if isinstance(input, dict):
+        return {byteify(key): byteify(value)
+                for key, value in input.iteritems()}
+    elif isinstance(input, list):
+        return [byteify(element) for element in input]
+    elif isinstance(input, unicode):
+        return input.encode('utf-8')
+    else:
+        return input
```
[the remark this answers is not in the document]
kbarkevich: Python 2 does not allow us to directly read the file with a `utf-8` encoding. If this was Python 3, I could open the file like so: `open(self.file_loc, 'r', encoding='utf-8')`. However, the `encoding` keyword does not exist in python 2, and thus it becomes necessary to comb through the data structure loaded from the json file and manually encode its strings. This function accomplishes that.

mh/quest_utils.py
```diff
@@ -10,7 +10,19 @@
 from mh.equipment_utils import create_arena_equipment_set
 
 
-class LocationType:
+class Enum(object):
+    def __getitem__(self, idx):
+        return getattr(self, idx)
```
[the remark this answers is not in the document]
kbarkevich: This will allow us to create quests programatically from strings. For example, instead of just using the encoded number `8` to refer to barroth in a quest .json file, we can instead use the string `barroth` and the `Enum` nature of the `Monsters` class will allow us to get the appropriate number from that string (i.e. `Monsters['barroth']`, which resolves to `8`.)

mh/constants.py
```diff
@@ -7,21 +7,26 @@
...
+def make_event_slot(event):
+    return (make_binary_event_quest(event), event)
+
+
+def get_event_slot(slot_num, temporal_slot):
+    return lambda quest_loader: quest_loader[temporal_slot][slot_num] if slot_num < len(quest_loader[temporal_slot]) else b'\0' * 0x4B4
+
+
+def is_arena_quest(flags):
+    return ((1 * 2**4) * 16**0) & flags > 0
```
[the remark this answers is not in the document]
kbarkevich: This was deliberately written as an equation to make it clear which exact bit is being checked. In this case, it's the bit with index 4 (so the 5th bit) of the byte with index 0 (so the first byte) in the flags. That bit is the arena quest flag.

mh/constants.py
kbarkevich: This logic has been moved there

mh/constants.py
kbarkevich: In this case, since the `slot_num` and `temporal_slot` need to be preserved, I actually needed to add them as optional parameters to the lambda function that default to themselves (see: https://stackoverflow.com/questions/938429/scope-of-lambda-functions-and-their-parameters). I have, however, split this line into several to make it shorter.

mh/constants.py
kbarkevich: This has been done

mh/constants.py
kbarkevich: Currently yes, since all the dynamically updating (in other words, callable instead of static) binaries have the binary loader passed in just in case it's needed (line 1140 of `pat.py`)

mh/constants.py
kbarkevich: It was not intended and has now been fixed, thanks for pointing it out

mh/pat.py
kbarkevich: Not all binaries currently come from the `binary_loader`, only the ones that have the potential to change from moment to moment. In other words yes, it could be moved (and probably should, once we have things like a rotating trading post), but since this PR is specifically about the event quests, I'll leave that to a future update.

mh/quest_utils.py
kbarkevich: No, it has been removed

mh/quest_utils.py
kbarkevich: This has been removed

mh/quest_utils.py
kbarkevich: Fixed

mh3fan:
Hello, first of all thank you for setting up this MH3SP project and for bringing this magnificent game back to life, I don't know if I'm in the right place to talk about this but a lot of important event quests are missing, let me allow myself so to inform you that I have just found "all" the event quests of the game at this link : (https://monsterhunter.fandom.com/wiki/MH3:_Event_Quests#Rage_Match) hoping that this will help you, (I could possibly translate the quests into French), unfortunately I cannot "yet" help you with the development because I am learning, therefore my skills will not be sufficient, in any case good luck to you and thank you again, hoping that the information will be transmitted ^^

PR #114, opened 2023-07-15
Start server worker thread when expecting work not before

InusualZ:
Currently on `dev` if we try to run the server with Python v3 without `LegacySSL=ON` we would hit hit this error:
image
Which would crash the main thread, the issue is that the Python VM would not terminate the process instead it would wait for all the threads to be _joined_. This is problematic because the worker threads are expecting work to be done because they are started in the constructor of the server class. This in turn make the process unkillable by means of CTRL+C

Now we defer starting the worker threads until we actually start serving the server.

### DIFF
--- a/mh/server.py
+++ b/mh/server.py
@@ -170,7 +170,6 @@ class BasicPatServer(object):
             )
             self.worker_queues.append(thread_queue)
             self.worker_threads.append(thread)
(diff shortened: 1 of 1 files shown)


sepalani (approved):
LGTM.

PR #115, opened 2023-07-15
Fix deadlocks when an error occurs early

sepalani:
A follow up of https://github.com/sepalani/MH3SP/pull/114.

This PR fixes deadlocks when exceptions occur early (ex: ports already used) since `__is_shut_down` wasn't set and `close` is waiting for the event.

I also changed `max_threads` _(which only auto-detect when set to zero, now, which is the default value)_ as we don't use negative numbers, AFAICT.


InusualZ (approved):
LGTM

PR #120, opened 2023-10-21
Patched OPN server to fix the SSL stall issue

kbarkevich:
This patch should fix the issue with the login server/OPN server going down at random periods. The issue was caused by foreign connections being established to the OPN server's port and then no data being transmitted, resulting in an eternal "SSL stall" while the server waits for the handshake to be completed. This blocked any further incoming connections.

### DIFF
--- a/opn_server.py
+++ b/opn_server.py
@@ -6,17 +6,27 @@
 
 from mh.pat import PatServer, PatRequestHandler
 
-from other.utils import server_base, server_main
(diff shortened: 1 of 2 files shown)


sepalani:
This approach doesn't work if the `UseSSL` INI option is used by another server. I posted an alternative approach here: https://github.com/sepalani/MH3SP/pull/121

sepalani:
Superseded by https://github.com/sepalani/MH3SP/pull/121.

PR #121, opened 2023-10-21
Workaround to prevent TLS handshakes to hang the server

sepalani:
Generic alternative to https://github.com/sepalani/MH3SP/pull/120.

### DIFF
--- a/mh/server.py
+++ b/mh/server.py
@@ -11,6 +11,7 @@ import struct
 import threading
 
 from mh.time_utils import Timer
+from other.utils import wii_ssl_wrap_socket
 
 try:
     # Python 3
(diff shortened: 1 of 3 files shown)


mh/server.py
```diff
@@ -144,8 +145,8 @@ class BasicPatServer(object):
     socket_type = socket.SOCK_STREAM
 
     def __init__(self, server_address, RequestHandlerClass, max_threads,
-                 bind_and_activate=True):
-        # type: (Tuple[str, int], BasicPatHandler, int, bool) -> None
+                 bind_and_activate=True, ssl_cert=None, ssl_key=None):
+        # type: (Tuple[str, int], BasicPatHandler, int, bool, str, str) -> None
```
InusualZ, line 149:
```py
# type: (Tuple[str, int], BasicPatHandler, int, bool, Optional[str], Optional[str]) -> None
```

Or (if you don't want to do ```from typing import Optional```)
```py
# type: (Tuple[str, int], BasicPatHandler, int, bool, str|None, str|None) -> None
```

InusualZ (changes requested):
As per the stackoverflow's [post](https://stackoverflow.com/a/68214507) that you linked. We have to timeout non-responsive TCP connections. I suppose that's handled by the previous `client_socket.settimeout` which is nice.

sepalani:
> As per the stackoverflow's [post](https://stackoverflow.com/a/68214507) that you linked. We have to timeout non-responsive TCP connections. I suppose that's handled by the previous `client_socket.settimeout` which is nice.

@InusualZ 
Yeah, we've already used the `settimeout` to handle ill-formed packets. If we find another way to handle them, we'd have to restore the previous timeout setting after calling the SSL wrap function. I also addressed your comment.

Code seems good to me, I still need to test in on real hardware to make sure it doesn't introduce regressions.

InusualZ (approved):

sepalani:
Tested and working on real hardware.

PR #133, opened 2025-07-08
Backport the event rotation branch to dev

sepalani:
This PR backports code from the branches used by the open beta (i.e. event-quest-rotation/event-quest-rotation-full).

It should also fix:
 - Starting standalone (OPN, LMP, FMP) server
 - Python2/3 compatibility

Regarding some of the modification done to the original commits:
 - Load mh.quest_utils.QuestLoader("event/quest_rotation.json") from the FMP server directly (i.e. remove parameter pollution)
 - Skip .bat scripts, *.ini, remote/alternative config
 - Skip central/cache/session code and its whitelist code
 - Skip the MySQL code
 - Skip code reverted to a previous state without reason, I assume they're probably from merge conflicts (e.g. comments removal, code de-linted, log messages removed, etc.)
 - Specialise sendReqConnection rather than creating patchCheck functions

Otherwise, I do plan follow-up PRs to address the TODO/backport comments. I think I'll prioritise the MySQL part first, then implement the rest.


InusualZ (approved):
Porting seem to be pretty straightforward. Most of the additions come from the json (events).

opn_server.py
```diff
@@ -16,7 +17,35 @@ class OpnServer(PatServer):
...
+
+            if address == 0x80000000:
+                self.game_id = to_str(data)  # type: ignore
+                if self.game_id == 'RMHE08':
+                    self.sendReqMemoryCheck(0x806308e8, 56)
+                elif self.game_id == 'RMHP08':
+                    self.sendReqMemoryCheck(0x806311a8, 56)
+                else:
+                    PatRequestHandler.sendReqConnection(self)
+            elif address in [0x806308e8, 0x806311a8]:
```
InusualZ, line 46: I saw this while reviewing the other PR. I think this can be converted into a general purpose check. Since if you are hosting a private server, you may want to check that all the player can play with each other. So maybe adding a NATNEG domain field in the config file and checking that it match it wouldn't bad

sepalani: Yeah, I agree and do plan to add that in a follow-up PR. It'd be especially useful since the Japanese version used a different one (IIRC) and that the domain can also be changed using a (Riivolution) patch.

PR #134, opened 2025-07-19
Backport MySQL backend to dev

sepalani:
This PR adds MySQL backend support for both python2/3.

Depends on https://github.com/sepalani/MH3SP/pull/133.

### DIFF
--- a/mh/database.py
+++ b/mh/database.py
@@ -4,6 +4,7 @@
 # SPDX-License-Identifier: AGPL-3.0-or-later
 """Monster Hunter database module."""
 
+import inspect
 import random
 import sqlite3
 import time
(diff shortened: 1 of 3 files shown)


mh/database.py
```diff
@@ -886,7 +886,190 @@ def delete_friend(self, capcom_id, friend_id):
...
+                " (id MEDIUMINT NOT NULL AUTO_INCREMENT,"
+                " capcom_id VARCHAR(6) NOT NULL,"
+                " friend_id VARCHAR(6) NOT NULL,"
+                " PRIMARY KEY (id),"
+                " UNIQUE friends_uniq_idx (capcom_id, friend_id));"
+            )
+
+    def populate_database(self):
+        with self.connection.cursor() as cursor:
+            cursor.execute("SELECT * FROM consoles")
```
InusualZ, line 936: I don't quite like that we are loading the entire table into memory when not all slot may be used. After we land this, this should be refactored into a more lazy operation.

InusualZ (approved):
Only one comment that can be addressed after all the backport is complete.

